PR #9, opened 2024-02-01
Feature flag schema update

mrm9084:
# Description

* Updates the Schema to use "FeatureManagement": { "FeatureFlags": {} }
* Updates to support provider refresh. If using Azure App Configuration provider, feature flags will auto update there value after a provider refresh is done.

# Contribution checklist:
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**


README.md
```diff
@@ -47,12 +47,18 @@ A Json file with the following format can be used to load feature flags.
 
 ```json
 {
-    "FeatureManagement": [
-        {
-            "id": "Alpha",
-            "enabled": "true"
-        }
-    ]
+    "FeatureManagement": {
```
jimmyca15, line 50:
```suggestion
    "feature_management": {
```

README.md
```diff
@@ -47,12 +47,18 @@ A Json file with the following format can be used to load feature flags.
...
+        "FeatureFlags": [
```
jimmyca15, line 51:
```suggestion
        "feature_flags": [
```

microsoft/featuremanagement/_featuremanager.py
```diff
@@ -41,6 +36,20 @@ def __init__(self, feature_flags, **kwargs):
                 raise ValueError("Custom filter must be a subclass of FeatureFilter")
             self._filters[filter.name] = filter
 
+    def _get_feature_flag(self, feature_flag_name):
+        feature_management = self._configuration.get(FEATURE_MANAGEMENT_KEY)
+        if not feature_management:
+            feature_management = self._configuration
```
rossgrambo, line 42: We were considering not having this fallback for the new schema- since no one has it yet we could force it to adhere to the schema definition fully.

Meaning C# will only work with Microsoft Feature Management schema if it's fully `"feature_management" { "feature_flags": { } }`. 

Not sure if we want to follow suit here.

mrm9084: I'm personally not a huge fan of removing it as it's so easy to support... I would rather have it work than not.

I can bring it up with Jimmy and the rest to the team.

jimmyca15: I would start without it. Yes it is easy to support, but more options can sometimes be more harmful than helpful. More decisions for users of what's the "right thing to do", more things to think about when maintaining.

Since Python has the option to go without root fallback I would avoid it.

jimmyca15: @rossgrambo wanted to add, in the context of Python, this is not 'new schema' since it's the only one Python will support.

mrm9084: Updated to return `None` if the schema isn't correct when getting the feature flag.

microsoft/featuremanagement/aio/_featuremanager.py
```diff
@@ -24,19 +21,13 @@ class FeatureManager(SyncFeatureManager):
...
     """
 
-    def __init__(self, feature_flags, **kwargs):
-        self._feature_flags = {}
-        feature_management = feature_flags.get(FEATURE_MANAGEMENT_KEY, feature_flags)
-        for feature_flag_json in feature_management.get(FEATURE_FLAG_KEY, feature_management):
-            feature_flag = FeatureFlag.convert_from_json(feature_flag_json)
-            self._feature_flags[feature_flag.name] = feature_flag
-
+    def __init__(self, configuraiton, **kwargs):
```
jimmyca15, line 24:
```suggestion
    def __init__(self, configuration, **kwargs):
```

microsoft/featuremanagement/_featuremanager.py
```diff
@@ -41,6 +36,20 @@ def __init__(self, feature_flags, **kwargs):
                 raise ValueError("Custom filter must be a subclass of FeatureFilter")
             self._filters[filter.name] = filter
 
+    def _get_feature_flag(self, feature_flag_name):
+        feature_management = self._configuration.get(FEATURE_MANAGEMENT_KEY)
+        if not feature_management:
+            return None
+        feature_flags = feature_management.get(FEATURE_FLAG_KEY)
```
jimmyca15, line 43: Shouldn't you validate this to be dictionary before calling `get`?

microsoft/featuremanagement/_featuremanager.py
```diff
@@ -25,14 +25,9 @@ class FeatureManager:
...
     """
 
-    def __init__(self, feature_flags, **kwargs):
-        self._feature_flags = {}
-        feature_management = feature_flags.get(FEATURE_MANAGEMENT_KEY, feature_flags)
-        for feature_flag_json in feature_management.get(FEATURE_FLAG_KEY, feature_management):
-            feature_flag = FeatureFlag.convert_from_json(feature_flag_json)
-            self._feature_flags[feature_flag.name] = feature_flag
-
+    def __init__(self, configuraiton, **kwargs):
```
rossgrambo, line 28:
```suggestion
    def __init__(self, configuration, **kwargs):
```

microsoft/featuremanagement/_featuremanager.py
```diff
@@ -25,14 +25,9 @@ class FeatureManager:
...
         self._filters = {}
+        self._configuration = configuraiton
```
rossgrambo, line 30:
```suggestion
        self._configuration = configuration
```

microsoft/featuremanagement/aio/_featuremanager.py
```diff
@@ -24,19 +21,13 @@ class FeatureManager(SyncFeatureManager):
...
-    def __init__(self, feature_flags, **kwargs):
-        self._feature_flags = {}
-        feature_management = feature_flags.get(FEATURE_MANAGEMENT_KEY, feature_flags)
-        for feature_flag_json in feature_management.get(FEATURE_FLAG_KEY, feature_management):
-            feature_flag = FeatureFlag.convert_from_json(feature_flag_json)
-            self._feature_flags[feature_flag.name] = feature_flag
-
+    def __init__(self, configuration, **kwargs):
         self._filters = {}
+        self._configuration = configuraiton
```
rossgrambo, line 26:
```suggestion
        self._configuration = configuration
```

samples/feature_flag_with_azure_app_configuraiton_sample.py
```diff
@@ -0,0 +1,43 @@
+from microsoft.featuremanagement import FeatureManager
```
rossgrambo, line 1: File name has the configuraiton misspelling

microsoft/featuremanagement/_featuremanager.py
```diff
@@ -41,6 +36,20 @@ def __init__(self, feature_flags, **kwargs):
...
+        feature_management = self._configuration.get(FEATURE_MANAGEMENT_KEY)
+        if not feature_management:
+            return None
+        feature_flags = feature_management.get(FEATURE_FLAG_KEY)
+        if not feature_flags:
+            return None
+
+        for feature_flag in feature_flags:
+            if feature_flag.get("id") == feature_flag_name:
+                return FeatureFlag.convert_from_json(feature_flag)
```
jimmyca15, line 49: It looks to me like every time someone calls `is_enabled` there is a conversion from JSON, is that right?

jimmyca15 (approved):

microsoft/featuremanagement/_featuremanager.py
jimmyca15: If so, let's track an item to cache that.

mrm9084: Yes, that because we need to support the provider refreshing, but can't guarantee that the customer gave us json as a string vs a json/Map object.

jimmyca15: We can do better. There should be some happy path out there where is_enabled is very cheap.

samples/feature_flag_with_azure_app_configuraiton_sample.py
```diff
@@ -0,0 +1,43 @@
...
+from azure.appconfiguration.provider import load, SettingSelector
+import os
+from sample_utilities import get_authority, get_audience, get_credential, get_client_modifications
+import json
+import os
+import sys
+from time import sleep
+
+endpoint = os.environ.get("APPCONFIGURATION_ENDPOINT_STRING")
+authority = get_authority(endpoint)
```
rossgrambo, line 12: Why include sample_utilities and all of these get_authority, get_audience, get_credential, get_client_modifications in this sample? I think it's a helpful sample for AAD, but it makes this sample seem heavy. Especially when the util functions have their splitting logic.

microsoft/featuremanagement/_featuremanager.py
mrm9084: How? We just have a pointer to the "configuration" object. How would we know if it has changed or not? I guess we could hash the whole configuration and check if it changes?

Otherwise we have no real way to tell. We could just use the json and not the object. We use to do that, but was changed to a PR comment.

samples/feature_flag_with_azure_app_configuraiton_sample.py
mrm9084: @rossgrambo, so you're just saying use a connection string?

samples/feature_flag_with_azure_app_configuration_sample.py
```diff
@@ -0,0 +1,45 @@
...
+
+endpoint = os.environ.get("APPCONFIGURATION_ENDPOINT_STRING")
+authority = get_authority(endpoint)
+audience = get_audience(authority)
+credential = get_credential(authority)
+kwargs = get_client_modifications()
+
+# Connecting to Azure App Configuration using AAD
+config = load(
+    endpoint=endpoint, credential=credential, feature_flag_enabled=True, feature_flag_refresh_enabled=True, **kwargs
```
jimmyca15, line 19: Does the configuration provider not take connection string directly?

samples/feature_flag_with_azure_app_configuraiton_sample.py
rossgrambo: Yeah, I think so. This could be valuable to offer as well- but maybe this one should be "feature_flag_with_**aad**_app_configuration_sample.py" and then offer a simpler one that just uses connection string. 

Or leave out the aad auth and link to the python provider's example for how to use aad.

samples/feature_flag_with_azure_app_configuration_sample.py
mrm9084: It can, this is AAD. I can make it use a connection string the same way.

samples/feature_flag_with_azure_app_configuraiton_sample.py
mrm9084: I want a runnable sample, switching to connection string.

microsoft/featuremanagement/_featuremanager.py
rossgrambo: Javascript will likely have the same concern. Whatever we land on should be applied to both.

samples/sample_utilities.py
```diff
@@ -0,0 +1,61 @@
+# ------------------------------------------------------------------------
```
rossgrambo, line 1: This file is no longer needed- right?

rossgrambo (approved):

microsoft/featuremanagement/_featuremanager.py
Eskibear: > Javascript will likely have the same concern. Whatever we land on should be applied to both.

So far in JavaScript FM design, JSON parsing happens in provider.
Calling `is_enabled` only searches array of flags by name, which is not expensive as I tested before.

> Performance of array.find() : linear search, the worst case is ~19ms per 1 million flags. For feature flag number < 10K (should be large enough for most cases), the time cost < 1ms. 
Tested on CPU: Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz

PR #13, opened 2024-02-27
Rename Package

mrm9084:
Changes from `microsoft.featuremanagement` to `featuremanagement`


docs/conf.py
```diff
@@ -5,9 +5,9 @@
 import sys
 import sphinx_rtd_theme
 
-sys.path.insert(0, os.path.abspath("../microsoft"))
```
rossgrambo, line 8: This doesn't seem like a 1:1 mapping like all the other changes. I assume the path before was "/microsoft/featuremanagement". Wouldn't this change be making it "featuremanagement/featuremanagement"?

featuremanagement/aio/_featuremanager.py
```diff
@@ -21,14 +22,14 @@ class FeatureManager(SyncFeatureManager):
...
     :param configuration: Configuration object
-    :type configuration: dict
+    :type configuration: Mapping
     :keyword feature_filters: Custom filters to be used for evaluating feature flags
     :paramtype feature_filters: list[FeatureFilter]
     """
 
     def __init__(self, configuration, **kwargs):
         self._filters = {}
-        if configuration is None or not isinstance(configuration, dict):
```
rossgrambo, line 31: Mapping change shouldn't be a part of this PR- but I recall why this is the case so it's fine with me.

docs/conf.py
mrm9084: ? This is the file path we just removed `microsoft` from the file path, nothing else changed. This is just pointing to where our code is for autodoc generation.

featuremanagement/aio/_featuremanager.py
mrm9084: The issue was that we missed updating async in the other PR and I didn't want it to get lost.

rossgrambo:
Does `sphinx-build -M html .\docs\ .\docs\` still work for you? I get 
```
WARNING: autodoc: failed to import module 'aio' from module 'featuremanagement'; the following exception was raised:
No module named 'featuremanagement'
WARNING: autodoc: failed to import module 'featuremanagement'; the following exception was raised:
No module named 'featuremanagement'
```

docs/conf.py
rossgrambo: Changing this to 
```python
sys.path.insert(0, os.path.abspath("../"))
```

Fixes the sphinx build and I makes sense to me

mrm9084: Solved with new install

rossgrambo (approved):

PR #14, opened 2024-03-05
Refresh Cache

mrm9084:
# Description
Implements a cache that supports refresh

# Contribution checklist:


featuremanagement/_featuremanager.py
```diff
@@ -67,7 +69,14 @@ def is_enabled(self, feature_flag_id, **kwargs):
         :return: True if the feature flag is enabled for the given context
         :rtype: bool
         """
-        feature_flag = self._get_feature_flag(feature_flag_id)
+        if self._copy is not self._configuration.get(FEATURE_MANAGEMENT_KEY):
+            self._cache = {}
+            self._copy = self._configuration.get(FEATURE_MANAGEMENT_KEY)
+        if not self._cache.get(feature_flag_id):
```
rossgrambo, line 75: Adding whitespace
```suggestion
        
        if not self._cache.get(feature_flag_id):
```

tests/test_feature_manager_refresh.py
```diff
@@ -21,8 +21,13 @@ def test_refresh(self):
         feature_manager = FeatureManager(feature_flags)
         assert feature_manager is not None
         assert feature_manager.is_enabled("Alpha")
+        assert feature_manager.is_enabled("Alpha")  # test cache
```
rossgrambo, line 24: I don't think this proves the cache is working. Can we check if the cache object contains a record instead? 

Code coverage might prove it- but I'm unable to see that.

rossgrambo:
This looks acceptable to me, but I want to ensure @jimmyca15 and @zhenlan are okay with it as well. There's no cache expiring and it's entirely dependent on the ref changing. If someone is using a local file, they can't simply update the dict unless they're also updating the ref of "feature_flags".

rossgrambo (approved):

PR #15, opened 2024-03-06
Feature Varaints - isEnabled

mrm9084:
# Description

Addes Variants, only supports variant override.

# Contribution checklist:
- [x] **CHANGELOG is updated for new features, bug fixes or other significant changes.**


featuremanagement/_featuremanager.py
```diff
@@ -8,7 +8,9 @@
 from ._featurefilters import FeatureFilter
 from ._models._feature_flag import FeatureFlag
```
rossgrambo, line 9: Could remove this since the new import includes it.

mrm9084: Didn't notice that, that's why the other PR is needed.

featuremanagement/_featuremanager.py
```diff
@@ -70,14 +72,78 @@ def __init__(self, configuration, **kwargs):
...
+                    return True
+                if variant.status_override == "Disabled":
+                    return False
+        return status
+
+    @staticmethod
+    def _is_targeted(context_id):
+        """Determine if the user is targeted for the given context"""
+        hashed_context_id = hashlib.sha256(context_id.encode()).hexdigest()
+        context_marker = abs(int(hashed_context_id, 16))
```
rossgrambo, line 114: Does this need to use the new endian aware method?

featuremanagement/_featuremanager.py
```diff
@@ -70,14 +72,78 @@ def __init__(self, configuration, **kwargs):
...
+                        return user_allocation.variant
+        if feature_flag.allocation.group:
+            groups = kwargs.get("groups")
+            if groups:
+                for group_allocation in feature_flag.allocation.group:
+                    for group in groups:
+                        if group in group_allocation.groups:
+                            return group_allocation.variant
+        if feature_flag.allocation.percentile:
+            user = kwargs.get("user", "")
```
rossgrambo, line 136: Don't need a default because the default is None, right?

featuremanagement/_featuremanager.py
mrm9084: Oh, yeah, this PR was created before we found this out.

featuremanagement/_featuremanager.py
mrm9084: The code below this will break if we don't have a String value. `""` is the assumed user if only a group is provided for allocation. Same if no user or group is provided?

featuremanagement/_featuremanager.py
```diff
@@ -104,14 +170,16 @@ def is_enabled(self, feature_flag_id, **kwargs):
...
-            return False
+            return FeatureManager._check_default_disabled_variant(feature_flag)
 
         feature_conditions = feature_flag.conditions
         feature_filters = feature_conditions.client_filters
 
         if len(feature_filters) == 0:
             # Feature flags without any filters return evaluate
-            return True
+            return FeatureManager._check_default_enabled_variant(feature_flag)
```
rossgrambo, line 180: This should be
```suggestion
            result = True
```

Because variant allocation lower down may override it. And the `_check_default_enabled_variant` function doesn't run the % based allocation.

featuremanagement/_models/_allocation.py
```diff
@@ -0,0 +1,193 @@
...
+                allocation._user.append(UserAllocation(**user_allocation))
+        if GROUP in json:
+            allocations = json.get(GROUP)
+            for group_allocation in allocations:
+                allocation._group.append(GroupAllocation(**group_allocation))
+        if PERCENTILE in json:
+            allocations = json.get(PERCENTILE)
+            for percentile_allocation in allocations:
+                allocation._percentile.append(PercentileAllocation.convert_from_json(percentile_allocation))
+        allocation._seed = json.get(SEED, "")
```
rossgrambo, line 53: Why does seed default to None in the init, but defaults to "" here? I would imagine it should stay None.

featuremanagement/_featuremanager.py
mrm9084: See updated code. Do you think it works now?

featuremanagement/_models/_allocation.py
mrm9084: It should never be none. Fixing the other location.

featuremanagement/_featuremanager.py
rossgrambo: Which code below? A None will stop us from entering the if statement like all the other `get()` calls above.

featuremanagement/_featuremanager.py
rossgrambo: The updated code should still just be `result["enabled"] = true`, otherwise it will check the default enabled variant, override to false, then later in the code check the disabled default and it could override back to true.

featuremanagement/_featuremanager.py
mrm9084: That None check isn't needed. That's why we have this issue.

featuremanagement/_models/_allocation.py
mrm9084: actually, the default should be "allocation\n<feature-name>", checked DotNet.

featuremanagement/_featuremanager.py
```diff
@@ -70,14 +73,87 @@ def __init__(self, configuration, **kwargs):
...
+        if feature_flag.allocation.group:
+            groups = kwargs.get("groups")
+            if groups:
+                for group_allocation in feature_flag.allocation.group:
+                    for group in groups:
+                        if group in group_allocation.groups:
+                            return group_allocation.variant
+        if feature_flag.allocation.percentile:
+            user = kwargs.get("user", "")
+            context_id = user + "\n" + feature_flag.name + "\n" + feature_flag.allocation.seed
```
mrm9084, line 137: What is expected here?
https://github.com/microsoft/FeatureManagement-Dotnet/blob/5c2af2c403c2f1098102a797abb83805379b7652/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs#L330
https://github.com/microsoft/FeatureManagement-Dotnet/blob/5c2af2c403c2f1098102a797abb83805379b7652/src/Microsoft.FeatureManagement/FeatureManager.cs#L620

In DotNet this value is `user\nhint`, where hint is the seed. but hint is value as the user defined value or "allocation\n{feature_name}", but wouldn't this cause all variants to return the same value if a seed is not provided.

mrm9084: I've corrected this, but the change is in the default value in allocaiton.

featuremanagement/_models/_allocation.py
```diff
@@ -0,0 +1,193 @@
...
+                allocation._user.append(UserAllocation(**user_allocation))
+        if GROUP in json:
+            allocations = json.get(GROUP)
+            for group_allocation in allocations:
+                allocation._group.append(GroupAllocation(**group_allocation))
+        if PERCENTILE in json:
+            allocations = json.get(PERCENTILE)
+            for percentile_allocation in allocations:
+                allocation._percentile.append(PercentileAllocation.convert_from_json(percentile_allocation))
+        allocation._seed = json.get(SEED, "allocation\n" + feature_name)
```
rossgrambo, line 53:
```suggestion
        allocation._seed = json.get(SEED, allocation._seed)
```

featuremanagement/_featuremanager.py
rossgrambo:
```suggestion
            context_id = user + "\n" + feature_flag.allocation.seed
```

mrm9084: I'm not sure what's going on here. Your suggested change is already what the change is...

rossgrambo (approved):

PR #16, opened 2024-03-08
Unknown Filter Error

mrm9084:
# Description

If a Feature flag has an unknown filter it should throw an exception

# Contribution checklist:
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**

### DIFF
--- a/featuremanagement/_featuremanager.py
+++ b/featuremanagement/_featuremanager.py
@@ -104,6 +104,8 @@ class FeatureManager:
                 else:
                     if self._filters[filter_name].evaluate(feature_filter, **kwargs):
                         return True
+            else:
+                raise ValueError("Feature flag {} has unknown filter {}".format(feature_flag_id, filter_name))
         # If this is reached, and true, default return value is true, else false
         return feature_conditions.requirement_type == REQUIREMENT_TYPE_ALL
 
--- a/featuremanagement/_featuremanager.py
+++ b/featuremanagement/_featuremanager.py
@@ -80,7 +80,7 @@ class FeatureManager:
             feature_flag = self._cache.get(feature_flag_id)
 
         if not feature_flag:
-            logging.warning("Feature flag {} not found".format(feature_flag_id))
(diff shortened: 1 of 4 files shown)


featuremanagement/_featuremanager.py
```diff
@@ -104,6 +104,8 @@ def is_enabled(self, feature_flag_id, **kwargs):
                 else:
                     if self._filters[filter_name].evaluate(feature_filter, **kwargs):
                         return True
+            else:
+                raise ValueError("Feature flag {} has unknown filter {}".format(feature_flag_id, filter_name))
```
rossgrambo, line 108: If we moved from `.format("")` to `f""` above, should we do it here as well?

tests/test_feature_manager.py
```diff
@@ -77,6 +78,22 @@ def test_feature_manager_creation_with_filters(self):
...
+        }
+        feature_manager = FeatureManager(feature_flags, feature_filters=[AllwaysOn(), AllwaysOff()])
+        assert feature_manager is not None
+        with pytest.raises(ValueError) as e_info:
+            feature_manager.is_enabled("Alpha")
+        assert e_info.type == ValueError
+        assert e_info.value.args[0] == "Feature flag Alpha has unknown filter UnknownFilter"
+        
 
 class AllwaysOn(FeatureFilter):
```
rossgrambo, line 98: Should `AllwaysOn` be `AlwaysOn`?

featuremanagement/_featuremanager.py
rossgrambo, line 108: Not required here, but in .NET we offer an option that doesn't throw when this happens. ( IgnoreMissingFeatureFilters ), we just log in when it happens. By default it throws though.

rossgrambo (approved):

featuremanagement/_featuremanager.py
mrm9084: Maybe a feature to add later, in Spring this falls under the "fail-fast" option, but such a thing doesn't exist or is atypical in Python.

PR #17, opened 2024-03-08
Test Validation

mrm9084:
* Adds action to run; pylint, black, and pytest on all PRs
* Splits FeatureFilter and it's Async version into 2 distinct classes to solve Pylint issues.
* Adds internal sdocs
* Pylint updates
* Fixes a bug where list feature flags didn't do anything
* Removed dead code

### DIFF
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -15,6 +15,7 @@ class FeatureFilter(ABC):
     def evaluate(self, context, **kwargs):
         """
         Determine if the feature flag is enabled for the given context
+        
         :param Mapping context: Context for the feature flag
         :paramtype context: Mapping
         """
@@ -23,6 +24,7 @@ class FeatureFilter(ABC):
     def name(self):
         """
         Get the name of the filter
+
         :return: Name of the filter, or alias if it exists
         :rtype: str
         """
@@ -32,6 +34,13 @@ class FeatureFilter(ABC):
 
     @staticmethod
     def alias(alias):
+        """
+        Decorator to set the alias for the filter
+
+        :param str alias: Alias for the filter
+        :return: Decorator
+        :rtype: callable
+        """
         def wrapper(cls):
             cls._alias = alias
             return cls
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -6,7 +6,7 @@
 from abc import ABC, abstractmethod
 
 
-class FeatureFilter(ABC):
+class FeatureFilter(ABC):  # pylint: disable=duplicate-code
     """
     Parent class for all feature filters
     """
@@ -27,7 +27,7 @@ class FeatureFilter(ABC):
 
         :return: Name of the filter, or alias if it exists
         :rtype: str
-        """
+        """  # pylint: disable=duplicate-code
         if hasattr(self, "_alias"):
             return self._alias
         return self.__class__.__name__
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -15,7 +15,7 @@ class FeatureFilter(ABC):
     def evaluate(self, context, **kwargs):
         """
         Determine if the feature flag is enabled for the given context
-        
+
         :param Mapping context: Context for the feature flag
         :paramtype context: Mapping
         """
@@ -41,8 +41,9 @@ class FeatureFilter(ABC):
         :return: Decorator
         :rtype: callable
         """
+
         def wrapper(cls):
-            cls._alias = alias # pylint: disable=protected-access
+            cls._alias = alias  # pylint: disable=protected-access
             return cls
 
         return wrapper
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -6,7 +6,7 @@
 from abc import ABC, abstractmethod
 
 
-class FeatureFilter(ABC):  # pylint: disable=duplicate-code
+class FeatureFilter(ABC):
     """
     Parent class for all feature filters
     """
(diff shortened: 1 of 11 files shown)


featuremanagement/_models/_feature_flag.py
```diff
@@ -15,60 +15,133 @@
...
+    Represents the conditions for a feature flag
+    """
+
+    def __init__(self):
+        self._requirement_type = REQUIREMENT_TYPE_ANY
+        self._client_filters = []
+
     @classmethod
-    def convert_from_json(cls, json):
+    def convert_from_json(cls, json_value):
```
albertofori, line 27: Do we need to validate that this is a Mapping/dict?

featuremanagement/aio/_defaultfilters.py
```diff
@@ -43,11 +26,12 @@ async def evaluate(self, context, **kwargs):
         :return: True if the current time is within the time window
         :rtype: bool
         """
-        return super().evaluate(context, **kwargs)
+        time_window_filter = SyncTimeWindowFilter()
+        return time_window_filter.evaluate(context, **kwargs)
```
albertofori, line 30: nit: can this be in-line?

mrm9084: We should actually probably not create a new instance every time.

featuremanagement/_featurefilters.py
```diff
@@ -23,17 +24,26 @@ def evaluate(self, context, **kwargs):
     def name(self):
         """
         Get the name of the filter
+
         :return: Name of the filter, or alias if it exists
         :rtype: str
-        """
+        """  # pylint: disable=duplicate-code
```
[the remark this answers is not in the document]
mrm9084: pylint detects that the code here and the async code is almost the same so it's flagging it. We don't really have a great way to reuse the code given where the async stuff is.

As for the two of them no idea why I need both of them, if you remove one other you get the duplicate-code message.

featuremanagement/_featuremanager.py
```diff
@@ -21,7 +20,40 @@
 FEATURE_FILTER_PARAMETERS = "parameters"
 
 
-class FeatureManager:
+def _get_feature_flag(configuration, feature_flag_name):
```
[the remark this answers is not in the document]
mrm9084: So, before we had the Async library extend the sync one, but pylint was throwing errors on that as it didn't like how we were overriding a sync method with an async one. It did work, but I was afraid that customers would get similar issues/warnings from pylint. My solution to the code duplication stuff was move the shared sync code out to be not at the class level and then I import the function into the sync code for use.

featuremanagement/_featurefilters.py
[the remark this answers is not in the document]
avanigupta: Looks like SDK has project level disables for cases like this:
https://github.com/Azure/azure-sdk-for-python/blob/5cff7c90a59a3c8745659ce209fe420a0d328594/pylintrc#L23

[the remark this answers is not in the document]
mrm9084: Added to pyproject.toml

featuremanagement/_featuremanager.py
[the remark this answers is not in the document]
mrm9084: I took a peak at there code, they use a `utils` module that they both import from. We could do the same. It's equivalent to what I'm doing here but in a separate file.

featuremanagement/_models/_feature_flag.py
```diff
@@ -15,60 +16,135 @@
...
+
+        :param json: JSON object
+        :type json: dict
+        :return: FeatureConditions
+        :rtype: FeatureConditions
+        """
         conditions = cls()
-        conditions._requirement_type = json.get(FEATURE_FILTER_REQUIREMENT_TYPE, REQUIREMENT_TYPE_ANY)
-        conditions._client_filters = json.get(FEATURE_FLAG_CLIENT_FILTERS, [])
+        if json_value is not None and not isinstance(json_value, Mapping):
```
rossgrambo, line 38:
```suggestion
        if json_value is not None or not isinstance(json_value, Mapping):
```

featuremanagement/_models/_feature_flag.py
```diff
@@ -15,60 +16,135 @@
...
-                raise ValueError("Feature flag {} is missing filter name.".format(feature_flag_id))
+                raise ValueError(f"Feature flag {feature_flag_id} is missing filter name.")
+
 
 class FeatureFlag:
+    """
+    Represents a feature flag
+    """
+
+    def __init__(self):
```
rossgrambo, line 77: I don't feel too strongly either way, but can we tell the linter to not require these init functions if we don't really use them?

mrm9084: We can, but either we need ignores on every method that does have the first use of a class variable, or we would have to set it as a global rule, see the other change I just made.

I actually don't mind them as they make it clear what default values are, though it will be more clear when we add Variant support where this gets a lot more complex.

featuremanagement/aio/_featurefilters.py
```diff
@@ -3,19 +3,47 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # -------------------------------------------------------------------------
-from abc import abstractmethod
-from .._featurefilters import FeatureFilter as FF
+from abc import ABC, abstractmethod
 
 
-class FeatureFilter(FF):
+class FeatureFilter(ABC):
```
rossgrambo, line 9: Does this need ABC as a param? Isn't it available without placing it here?

mrm9084: `ABC` is marking it as an `Abstract Class`

rossgrambo (approved):

rossgrambo:
lgtm assuming we fix the failing tests

PR #19, opened 2024-03-18
Fixing Targeting Issue + Tests

mrm9084:
# Description

Fixes an issue with Targeting where its results didn't match DotNet. DotNet is Little endien and unsigned.

Added test file to test all feature flags with expected results.

### DIFF
--- a/featuremanagement/_defaultfilters.py
+++ b/featuremanagement/_defaultfilters.py
@@ -83,10 +83,7 @@ class TargetingFilter(FeatureFilter):
             return True
 
         hashed_context_id = hashlib.sha256(context_id.encode()).digest()
-
-        byte_array = bytearray(hashed_context_id)
-        decimal_values = [int(byte) for byte in byte_array[:4]]
-        context_marker = int.from_bytes(decimal_values, byteorder="little", signed=False)
+        context_marker = int.from_bytes(hashed_context_id[:4], byteorder="little", signed=False)
 
         percentage = (context_marker / (2**32 - 1)) * 100
         return percentage < rollout_percentage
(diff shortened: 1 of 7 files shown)


featuremanagement/_defaultfilters.py
```diff
@@ -11,7 +11,7 @@
 
 from ._featurefilters import FeatureFilter
 
-FEATURE_FLAG_NAME_KEY = "name"
+FEATURE_FLAG_NAME_KEY = "feature_name"
```
amerjusupovic, line 14: I'm not super familiar with this library, but why is this key getting changed? Did the `context` variable structure change?

featuremanagement/_defaultfilters.py
```diff
@@ -82,14 +82,19 @@ def _is_targeted(context_id, rollout_percentage):
         if rollout_percentage == 100:
             return True
 
-        hashed_context_id = hashlib.sha256(context_id.encode()).hexdigest()
-        context_marker = abs(int(hashed_context_id, 16))
-        percentage = (context_marker / (2**256 - 1)) * 100
+        hashed_context_id = hashlib.sha256(context_id.encode()).digest()
 
+        byte_array = bytearray(hashed_context_id)
+        decimal_values = [int(byte) for byte in byte_array[:4]]
```
rossgrambo, line 88: This line doesn't do anything as far as I can tell. We can just pass `byte_array[:4]` into the next line instead.

mrm9084: I think that section got added as debug steps. Removing.

rossgrambo (approved):

featuremanagement/_defaultfilters.py
amerjusupovic: Gotcha, thanks

amerjusupovic (approved):

PR #20, opened 2024-03-22
Validation tests

mrm9084:
# Description

Updates the validation tests to be across multiple files.


tests/validation_tests/TargetingFilter.tests.json
```diff
@@ -0,0 +1,116 @@
+{
+    "1": {
```
zhiyuanliang-ms, line 2: Representing the list as set of properties is not a good practice.

https://github.com/Azure/AppConfiguration/pull/890#discussion_r1538503536

featuremanagement/_defaultfilters.py
```diff
@@ -92,9 +92,7 @@ def _target_group(self, target_user, target_group, group, feature_flag_name):
         group_rollout_percentage = group.get(ROLLOUT_PERCENTAGE_KEY, 0)
         if not target_user:
             target_user = ""
-        audience_context_id = (
-            target_user + "\n" + target_group + "\n" + feature_flag_name + "\n" + group.get(FEATURE_FILTER_NAME_KEY, "")
-        )
+        audience_context_id = target_user + "\n" + feature_flag_name + "\n" + target_group
```
albertofori, line 95: Just curious, is this to be consistent with our other providers?

featuremanagement/_models/_feature_flag.py
```diff
@@ -115,9 +115,9 @@ def variants(self):
 
     def _validate(self):
         if not isinstance(self._id, str):
-            raise ValueError("Feature flag id field must be a string.")
+            raise ValueError(f"Invalid setting 'id' with value '{self._id}' for feature '{self._id}'.")
```
albertofori, line 118: If the class of this instance does not have a __str__ or __repr__ defined, the value could appear weird in the error message. we say something along these lines:
 `Invalid type for setting 'id'. 'str'  expected but {type(self._id).__name__} given`

featuremanagement/_defaultfilters.py
mrm9084: Yeah, this is a bug found by the linked tests. https://github.com/microsoft/FeatureManagement-Dotnet/blob/709bbebd89e2923f820f75e651fd39778df26cbe/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs#L119

featuremanagement/_models/_feature_flag.py
mrm9084: This change was to start moving us towards consistent error messages between the libraries, this is the exact way DotNet does the message.

featuremanagement/_defaultfilters.py
albertofori: Got it, thanks!

albertofori (approved):

featuremanagement/_models/_feature_flag.py
rossgrambo: I like the idea of being consistent, but in dotnet- I don't think this error will ever happen on name/id and I think the wording is weird here for the str check.

This error will only happen if i'ts not a string, but when logging it would say something like:
`Invalid setting 'id' with value 'null' for feature 'null'` - which is kind of clear, but number and object could be weird: `Invalid setting 'id' with value '1' for feature '1'`, because string:"1" is a valid name but number:1 is not. 

I prefer something closer to @albertofori 's suggestion.

rossgrambo (approved):

featuremanagement/_models/_feature_flag.py
mrm9084: This isn't a great example as 1 is valid as python is typeless.

I can update this in a sepreate PR, I can change to what Albert mentioned, it's the change bellow this one that is 1 to 1 with Dot Net.

PR #23, opened 2024-04-01
Feature Variant Readme

mrm9084:
Updates readme for variants.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ class MyCustomFilter(FeatureFilter):
 
 When new features are added to an application, there may come a time when a feature has multiple different proposed design options. A common solution for deciding on a design is some form of A/B testing, which involves providing a different version of the feature to different segments of the user base and choosing a version based on user interaction. In this library, this functionality is enabled by representing different configurations of a feature with variants.
 
-Variants enable a feature flag to become more than a simple on/off flag. A variant represents a value of a feature flag that can be a string, a number, a boolean, or even a configuration object. A feature flag that declares variants should define under what circumstances each variant should be used, which is covered in greater detail in the [Allocating Variants](./README.md#AllocatingVariants) section.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -268,6 +268,164 @@ class MyCustomFilter(FeatureFilter):
     ...
 ```
 
+### Feature Variants
+
+When new features are added to an application, there may come a time when a feature has multiple different proposed design options. A common solution for deciding on a design is some form of A/B testing, which involves providing a different version of the feature to different segments of the user base and choosing a version based on user interaction. In this library, this functionality is enabled by representing different configurations of a feature with variants.
+
+Variants enable a feature flag to become more than a simple on/off flag. A variant represents a value of a feature flag that can be a string, a number, a boolean, or even a configuration object. A feature flag that declares variants should define under what circumstances each variant should be used, which is covered in greater detail in the [Allocating Variants]() section.
```
amerjusupovic, line 275: No link for `Allocating Variants`

README.md
```diff
@@ -268,6 +268,164 @@ class MyCustomFilter(FeatureFilter):
...
+feature_manager = FeatureManager(feature_flags)
+
+variant = feature_manager.get_variant("FeatureU")
+
+my_configuration = variant.configuration
+
+# Do something with the resulting variant and its configuration
+```
+
+Once a variant is retrieved, the configuration of a variant can be used directly as JSON from the variant's `configuration` property.
```
amerjusupovic, line 301: Should we put something here to demonstrate how the configuration could be used as JSON?

mrm9084: Moved this up to the section above, then also added a extra section to the sample with json.

README.md
```diff
@@ -268,6 +268,164 @@ class MyCustomFilter(FeatureFilter):
...
+"variants": [
+    { 
+        "name": "On"
+    },
+    { 
+        "name": "Off",
+        "status_override": "Disabled"
+    }    
+],
+"enabled_ror": [
```
amerjusupovic, line 420:
```suggestion
"enabled_for": [ 
```

amerjusupovic (approved):

PR #24, opened 2024-04-17
Telemetry Support

mrm9084:
# Description

Adds Telemetry support + changes from design meeting
* user_id without parameter
* Added Targeting_Context


featuremanagement/_featuremanager.py
```diff
@@ -64,6 +64,8 @@ class FeatureManager:
     :type configuration: Mapping
     :keyword feature_filters: Custom filters to be used for evaluating feature flags
     :paramtype feature_filters: list[FeatureFilter]
+    :keyword on_feature_evaluated: Callback function to be called when a feature flag is evaluated
```
rossgrambo, line 67: I think we may want to rethink this callback function. This makes sense with the publisher idea- but OpenTelemetry's whole goal is to stop custom callbacks like this. It might make sense to use their api to push an event here instead and let an exporter handle pushing the event to app insights/whereever. 

In .net- I'm working on a design doc now to adjust to this approach.

featuremanagement/_models/_targeting_context.py
```diff
@@ -0,0 +1,19 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
```
zhiyuanliang-ms, line 6: Is the convention of this repo to add an emtpy line between copyright and imports?

featuremanagement/_models/_evaluation_event.py
```diff
@@ -0,0 +1,23 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# -------------------------------------------------------------------------
+from dataclasses import dataclass
+
+
+@dataclass
+class EvaluationEvent:
+    """
+    Represents an evaluation event
```
zhiyuanliang-ms, line 12: Represents a feature flag evaluation event

featuremanagement/_models/_feature_flag.py
```diff
@@ -113,6 +117,16 @@ def variants(self):
         """
         return self._variants
 
+    @property
+    def telemetry(self):
+        """
+        Get the telemetry for the feature flag
```
zhiyuanliang-ms, line 123: This looks ambiguity to me.

How about "Get the telemetry configuration for the feature flag"?

featuremanagement/_models/_telemetry.py
```diff
@@ -0,0 +1,16 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# -------------------------------------------------------------------------
+from dataclasses import dataclass, field
+
+
+@dataclass
+class Telemetry:
+    """
+    Represents the telemetry for a feature flag
```
zhiyuanliang-ms, line 12:
```suggestion
    Represents the telemetry configuration for a feature flag
```

featuremanagement/_models/_targeting_context.py
```diff
@@ -0,0 +1,19 @@
...
+from typing import NamedTuple, List
+
+
+class TargetingContext(NamedTuple):
+    """
+    Represents the context for targeting a feature flag.
```
zhiyuanliang-ms, line 12: Would you mind standardizing the punctuation in the comments throughout the repo? It would be great if they could all end with a period or not.

featuremanagement/_models/_variant_assignment_reason.py
```diff
@@ -0,0 +1,19 @@
+# ------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# -------------------------------------------------------------------------
+from enum import Enum
+
+
+class VaraintAssignmentReason(Enum):
```
zhiyuanliang-ms, line 9: Would you mind adding some comments to explain each assignment reason?

You can refer to the .NET feature management.

/// <summary>
/// Variant allocation did not happend. No variant is assigned.
/// </summary>
None,

/// <summary>
/// The default variant is assigned when a feature flag is disabled.
/// </summary>
DefaultWhenDisabled,

/// <summary>
/// The default variant is assigned because of no applicable user/group/percentile allocation when a feature flag is enabled.
/// </summary>
DefaultWhenEnabled,

/// <summary>
/// The variant is assigned because of the user allocation when a feature flag is enabled.
/// </summary>
User,

/// <summary>
/// The variant is assigned because of the group allocation when a feature flag is enabled.
/// </summary>
Group,

/// <summary>
/// The variant is assigned because of the percentile allocation when a feature flag is enabled.
/// </summary>
Percentile

zhiyuanliang-ms:
@mrm9084 
Please refer to this PR: https://github.com/microsoft/FeatureManagement-Dotnet/pull/290

The variant assignment reason should be VariantAssignmentReason.NONE only when there is no Variant section in feature flag configuration. Discussion can be found [here](https://github.com/microsoft/FeatureManagement-Dotnet/pull/290#discussion_r1429380159)

BTW, I also didn't find the VariantAssignmentReason.DEFAULT_WHEN_ENABLED and DEFAULT_WHEN_DISABLED are used anywhere. Please refer to the .NET feature management.

zhiyuanliang-ms:
The overall variant assign reason should be:

if there is no Variants, the reason is "None"

if allocation is none/empty, which can be considered as none/empty "DefaultWhenEnabled" and "DefaultWhenDisabled"
even if we get null/none variant in this case, the reason should be "DefaultWhenEnabled" or "DefaultWhenDisabled"

if allocation is not none and we hit some allocation rule and get a variant, the reason should be that allocation rule name.

if no allocation rule is hit and we get null/none variant,  the reason should be "DefaultWhenEnabled" or "DefaultWhenDisabled"

The status_override of variant will not affect the variant assignment reason, because it takes place after the variant is assigned.


@zhenlan @jimmyca15 Please confirm whether this is correct.

tests/test_send_telemetry_appinsights.py
```diff
@@ -0,0 +1,96 @@
...
+from unittest.mock import patch
+import pytest
+from featuremanagement import EvaluationEvent, FeatureFlag, Variant, VaraintAssignmentReason
+import featuremanagement.azuremonitor._send_telemetry
+
+
+@pytest.mark.usefixtures("caplog")
+class TestSendTelemetryAppinsights:
+
+    def test_send_telemetry_appinsights(self):
```
zhiyuanliang-ms, line 19: I think we'd better have integration test like [this](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/tests/Tests.FeatureManagement/FeatureManagementTest.cs#L1336)

We call IsEnabled/GetVariant on a feature flag and we check the telemetry.

featuremanagement/_models/_targeting_context.py
zhiyuanliang-ms: The same issue also occurs in other files.

featuremanagement/_models/_variant_assignment_reason.py
zhiyuanliang-ms, line 9:
```suggestion
class VariantAssignmentReason(Enum):
```

featuremanagement/_models/_targeting_context.py
mrm9084: We are using the Python https://azure.github.io/azure-sdk/python_design.html, which includes using the black formatter and pylint. None of these specifies a style for this.

tests/test_send_telemetry_appinsights.py
mrm9084: These are unit tests, not integration tests.

mrm9084:
> The overall variant assign reason should be:
> 
> if there is no Variants, the reason is "None"
> 
> if allocation is none/empty, which can be considered as none/empty "DefaultWhenEnabled" and "DefaultWhenDisabled" even if we get null/none variant in this case, the reason should be "DefaultWhenEnabled" or "DefaultWhenDisabled"
> 
> if allocation is not none and we hit some allocation rule and get a variant, the reason should be that allocation rule name.
> 
> if no allocation rule is hit and we get null/none variant, the reason should be "DefaultWhenEnabled" or "DefaultWhenDisabled"
> 
> The status_override of variant will not affect the variant assignment reason, because it takes place after the variant is assigned.
> 
> https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L293
> 
> @zhenlan @jimmyca15 Please confirm whether this is correct.

So, basically the default value of reason should be `None` unless otherwise specified? I think the code is as is because no telemetry is sent when the reason is set to `None` so I ended up not setting a value.

tests/test_send_telemetry_appinsights.py
zhiyuanliang-ms: Can we add some integration test?

zhiyuanliang-ms:
> So, basically the default value of reason should be None unless otherwise specified? I think the code is as is because no telemetry is sent when the reason is set to None so I ended up not setting a value

Sorry, I just updated my previous comments. I didn't explain it clearly. 

Only when there is no `Variants` in the feature flag configuration, we will set variant assignment reason as `None` and it will be sent to the telemetry as a "None" string. So, I think you may remove the "NONE" in the VariantAssignmentReason enum in this PR because it will never be used. If you choose to set the reason of evaluation event as `None` instead of VariantAssignmentReason.NONE.

.

tests/test_send_telemetry_appinsights.py
mrm9084: I don't think any of our libraries besides Spring do actually integration tests. We would need to setup stores and resources to collect the telemetry.

mrm9084:
> > So, basically the default value of reason should be None unless otherwise specified? I think the code is as is because no telemetry is sent when the reason is set to None so I ended up not setting a value
> 
> Sorry, I just updated my previous comments. I didn't explain it clearly.
> 
> Only when there is no `Variants` in the feature flag configuration, we will set variant assignment reason as `None` and it will be sent to the telemetry as a "None" string. So, I think you may remove the "NONE" in the VariantAssignmentReason enum in this PR because it will never be used. If you choose to set the reason of evaluation event as `None` instead of VariantAssignmentReason.NONE.
> 
> .

We should try and be consistent in what we do between the libraries. As customers can access the `EvaluationEvent`. I'll just try summing up what I'm thinking.

By default, the value is `VariantAssignmentReason.NONE`.

But if the feature flag has any `Variant`s then the `VaraintAssignmentReason` should be set to something else while we process the feature flag.

Implementation wise. Both Python and Dot Net use the `EvaluationEvent` as a main object in the evaluation logic and the only difference between `is_enabled` and `get_variant` is how it is read at the end.

tests/test_send_telemetry_appinsights.py
zhiyuanliang-ms: Ok. I am just thinking there should be some testcases to test whether the feature evaluation process will produce the correct variant assignment reason.

featuremanagement/_models/_variant_assignment_reason.py
```diff
@@ -0,0 +1,19 @@
...
+# -------------------------------------------------------------------------
+from enum import Enum
+
+
+class VariantAssignmentReason(Enum):
+    """
+    Represents an assignment reason.
+    """
+
+    NONE = "NONE"
```
zhiyuanliang-ms, line 14: Should it be "None" since other reasons are all Pascal case

featuremanagement/aio/_featuremanager.py
```diff
@@ -26,6 +27,8 @@ class FeatureManager:
 
     :param Mapping configuration: Configuration object.
     :keyword list[FeatureFilter] feature_filters: Custom filters to be used for evaluating feature flags.
+    :keyword Callable[EvaluationEvent] on_feature_evaluated: Callback function to be called when a feature flag is
```
zhiyuanliang-ms, line 30: Is there any discussion about this? Currently, `on_feature_evaluated` will only take effect when featureflag.telemetry.enabled == true. This looks weird to me because the callback function is called "on feature evaluated", it's supposed to be called each time the feature flag is evaluated.

featuremanagement/aio/_featuremanager.py
```diff
@@ -181,7 +189,14 @@ async def is_enabled(self, feature_flag_id, *args, **kwargs):
         :rtype: bool
         """
         targeting_context = self._build_targeting_context(args)
-        return (await self._check_feature(feature_flag_id, targeting_context, **kwargs)).enabled
+        result = await self._check_feature(feature_flag_id, targeting_context, **kwargs)
+        if self._on_feature_evaluated and result.feature.telemetry.enabled:
+            result.user = kwargs.get("user", "")
```
zhiyuanliang-ms, line 194: Here is different from the implementation in async feature manager where you set result.user = targeting_context.user_id

zhiyuanliang-ms:
@mrm9084 Would you mind adjusting the order of methods by placing all public methods in front of private methods? I cannot put comment on those lines. I think "public" methods like `is_enabled`, `get_variant` should be placed in front of the "private" methods like `_variant_name_to_variant`

featuremanagement/aio/_featuremanager.py
mrm9084: @zhiyuanliang-ms, something went wrong with the merge. If you look at the last few commits I've been trying to fix them. But thanks for pointing this one out.

featuremanagement/aio/_featuremanager.py
mrm9084: The `on_feature_evaluated` was the named that was decided on during the design review meeting. I understand it's a bit weird but coming up with a name that fully describes what it does while keeping a reasonable keyword length is a challenge. I'm open up to hear options.

As is, this will go out in Beta for a while before GA, so we'll have time to change it.

mrm9084:
> @mrm9084 Would you mind adjusting the order of methods by placing all public methods in front of private methods? I cannot put comment on those lines. I think "public" methods like `is_enabled`, `get_variant` should be placed in front of the "private" methods like `_variant_name_to_variant`

I'll do this for new things. I don't want to move around old things as it will break the diff making it hard to review. I'll make a separate PR if there are more things to modify.

featuremanagement/aio/_featuremanager.py
```diff
@@ -108,26 +112,30 @@ def _assign_variant(self, feature_flag, targeting_context):
         :param TargetingContext targeting_context: Targeting context.
         :return: Variant name.
         """
+        evaluation_event = EvaluationEvent(feature_flag=feature_flag)
```
zhiyuanliang-ms, line 115: I think there should be only one EvaluationEvent instance during one feature evaluation which is the instance you created in the beginning of `_check_feature` method.

And all methods should modify that instance in place instead of returning a new instance or the EvaluationEvent instance passed in as a parameter.

zhiyuanliang-ms (changes requested):
I think there should be only one EvaluationEvent instance during one feature evaluation which is the instance you created in the beginning of _check_feature method.

And all methods should take that instance as a parameter and modify it in place instead of returning a new instance or the EvaluationEvent instance passed in as a parameter.

featuremanagement/_featuremanager.py
```diff
@@ -341,7 +360,7 @@ def _check_feature(self, feature_flag_id, targeting_context, **kwargs):
             evaluation_event.feature = feature_flag
             return evaluation_event
 
-        evaluation_event = self._check_feature_filters(feature_flag, targeting_context, **kwargs)
+        evaluation_event = self._check_feature_filters(feature_flag, evaluation_event, targeting_context, **kwargs)
```
zhiyuanliang-ms, line 363: Eventually, feature_flag will be set to evaluation_event.feature. And feature_flag will not be modified in `_check_feature_filters`, let's set `evaluation_event.feature = featuer_flag` in advance and remove `feature_flag` from the parameters of `_check_feature_filters`.

Currently, EvaluationEvent.Feature is never used.

zhiyuanliang-ms (changes requested):
Eventually, feature_flag will be set to evaluation_event.feature. And feature_flag will not be modified in _check_feature_filters, let's set evaluation_event.feature = featuer_flag in advance and remove feature_flag from the parameters of _check_feature_filters.

Currently, EvaluationEvent.Feature is never used.

featuremanagement/_featuremanager.py
```diff
@@ -100,9 +103,11 @@ def _check_default_disabled_variant(feature_flag):
         """
         if not feature_flag.allocation:
             return EvaluationEvent(enabled=False)
-        return FeatureManager._check_variant_override(
+        evaluation_event = FeatureManager._check_variant_override(
             feature_flag.variants, feature_flag.allocation.default_when_disabled, False
         )
+        evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED
```
zhiyuanliang-ms, line 109: When feature_flag.allocation is None and evaluation_event.enabled == false, the assigned variant should be `None` and reason should be "DefaultWhenDisabled". @jimmyca15 [ref](https://github.com/microsoft/FeatureManagement-Dotnet/pull/290#discussion_r1430793880)

Currently, if not feature_flag.allocation, this code path will be skipped at line 104 and the assignment reason will be set to "None".


Please refer to this [comment](https://github.com/microsoft/FeatureManagement-Python/pull/24#issuecomment-2106960349) and this [PR](https://github.com/microsoft/FeatureManagement-Dotnet/pull/290) for details.

zhiyuanliang-ms (changes requested):
When feature_flag.allocation is None and evaluation_event.enabled == false, the assigned variant should be `None` and reason should be "DefaultWhenDisabled". @jimmyca15 [ref](https://github.com/microsoft/FeatureManagement-Dotnet/pull/290#discussion_r1430793880)

Currently, if not feature_flag.allocation, this code path will be skipped at line 104 and the assignment reason will be left to None.


Please refer to this [comment](https://github.com/microsoft/FeatureManagement-Python/pull/24#issuecomment-2106960349) and this [PR](https://github.com/microsoft/FeatureManagement-Dotnet/pull/290) for details.

zhiyuanliang-ms:
The ideal code path of `_check_feature` method should be 

0. Create an `EvaluationEvent`
1. Get `FeatureFlag` from cache or configuration
2. Attach the `FeatureFlag` to EvaluationEvent
3. Call `_check_feature_filter` but only pass the `FeatureFlag` and feature filter evaluation context to it
4. Get the result of `_check_feature_filter` and attach the result to `EvaluationEvent.enabled`
5. Then we will assign the variant by calling `_assign_allocation`.

def _assign_allocation(evaluation_event, targeting_context) -> None:
    feature_flag = evalutation
    if feature_flag.variants:
        if not feature_flag.Allocation:
            ### evaluation_event.variant will be left to None
            if evaluation_event.enabled:
                evaluation_event.reason = "DefaultWhenEnabled"
            else:
                evaluation_event.reason = "DefaultWhenDisabled"
        elif evaluation_event.enabled == false:
            evaluation_event.variant = feature_flag.allocation.default_when_disabled
            evaluation_event.reason = "DefaultWhenDisabled"
        else:
            ### assign variant and set evaluation.variant and reason(User/Group/Percentile)
            _assign_variant(evaluation_event, targeting_context)

            if not evaluation_event.variant:
                evaluation_event.variant = feature_flag.allocation.default_when_enabled
                evaluation_event.reason = "DefaultWhenEnabled"

        if evaluation.variant and feature_flag.enabled:
            variant = _variant_name_to_variant(evaluation_event.variant)
            evaluation_event.enabled = variant.status_override

        return

    else:
        ### evaluation_event.variant and reason will be left to None
        return

featuremanagement/aio/_featuremanager.py
mrm9084: This should be updated everywhere

featuremanagement/_featuremanager.py
mrm9084: I think we are including this as part of the schema as we provide the customer the evaluation event if they want. DotNet does the same thing https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureDefinition.cs. @rossgrambo?

mrm9084:
> I don't think EvaluatonEvent.variant should store `VariantReference`. It should just store variant name. EvaluationEvent.feature has the whole feature flag definition, you can retrieve the variant reference from it.

`EvaluationEvent.variant` is the assigned variant. While the feature flag has all variants.

featuremanagement/_featuremanager.py
mrm9084: nvm, figured out what you meant. Should be updated.

featuremanagement/_featuremanager.py
mrm9084: This should be fixed now.

featuremanagement/_featuremanager.py
```diff
@@ -230,6 +247,9 @@ def is_enabled(self, feature_flag_id, *args, **kwargs):
         targeting_context = self._build_targeting_context(args)
 
         result = self._check_feature(feature_flag_id, targeting_context, **kwargs)
+        if self._on_feature_evaluated and result.feature.telemetry.enabled:
```
Eskibear, line 250: Maybe consider to change this condition to the following only, and leave the `feature.telemetry.enabled` check into the impl of `push_telemetry`?
```
if self._on_feature_evaluated
```

Reason:
- `on_feature_evaluated` is a callback function when ff is evaluated. It has nothing to do with what's inside a feature flag, and it doesn't have to be publishing telemetry. So there is no reason we check the evaluation result to determine whether to execute the callback or not.
- `publish_telemetry` is dedicated for telemetry. And it's reasonable that if feature.telemetry.enabled is false we don't send telemetry.

mrm9084: line 250 is checking if `_on_feature_evaluated` is not `None` this isn't actually calling the function.

Eskibear: I didn't mean that. I mean line 250 should not check `result.feature.telemetry.enabled` because the following block is to call `_on_feature_evaluated`, which doesn't have to be something related to "telemetry". (line 252)

featuremanagement/_featuremanager.py
```diff
@@ -283,31 +306,32 @@ def _check_feature_filters(self, feature_flag, targeting_context, **kwargs):
...
+        if feature_flag.variants:
+            if not feature_flag.allocation:
+                if evaluation_event.enabled:
+                    evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+                    return
+                evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED
+                return
+            if not evaluation_event.enabled:
+                FeatureManager._check_default_disabled_variant(evaluation_event)
+                evaluation_event.variant = self._variant_name_to_variant(
```
zhiyuanliang-ms, line 321: We can add a shortcut in `_variant_name_to_variant` when `variant_name` is None. For example, if `feature_flag.allocation.default_when_enabled` is `None` here, the for loop in Line 206 can be skipped.

featuremanagement/_featuremanager.py
```diff
@@ -283,31 +306,32 @@ def _check_feature_filters(self, feature_flag, targeting_context, **kwargs):
...
+                    feature_flag, feature_flag.allocation.default_when_enabled
+                )
+                return
+
+            variant_name = self._assign_variant(targeting_context, evaluation_event)
```
zhiyuanliang-ms, line 326: In `_assign_variant` method, `evaluation_event.reason` is set. Why not set `evaluation_event.variant` as well? 
It is weird to seperate them.

featuremanagement/_featuremanager.py
rossgrambo: `_on_feature_evaluated` is the callback customers setup for the purpose of telemetry. So the idea from the design was that `_on_feature_evaluated` should not be called on flags that don't have telemetry enabled. 

But I do agree that the naming sounds like this shouldn't be limited to telemetry. I think we should either:
1. Rename the callback to something along the lines of "telemetry_callback" 
2. Always call the callback- and let that function handle the `.telemetry.enabled` check.

I prefer option 2, and we add the telemetry check at the start of [`def publish_telemetry`](https://github.com/microsoft/FeatureManagement-Python/pull/24/files/7297743600d12819618130f58fa05fc87d962049#diff-e599fecd01f487185b6b19104e39d7a20d310657896ddeee742af944b129945aR43)

featuremanagement/opentelemetry/__init__.py
```diff
@@ -0,0 +1,13 @@
+# ------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# -------------------------------------------------------------------------
+
```
zhiyuanliang-ms, line 6: Could you remove this empty line please?

featuremanagement/opentelemetry/_send_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+
+FEATURE_NAME = "FeatureName"
+ENABLED = "Enabled"
+TARGETING_ID = "TargetingId"
+VARIANT = "Variant"
+REASON = "VariantAssignmentReason"
+
+EVENT_NAME = "FeatureEvaluation"
+
+_event_logger.propagate = False
```
zhiyuanliang-ms, line 35: What's the purpose of this line?

featuremanagement/opentelemetry/_send_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        if not _FeatureManagementEventsExtension._initialized:
+            _event_logger.addHandler(LoggingHandler())
+            _event_logger.setLevel(INFO)
+            _FeatureManagementEventsExtension._initialized = True
+
+
+def track_event(event_name, user, event_properties=None):
+    """
+    Track an event with the specified name and properties.
```
zhiyuanliang-ms, line 54:
```suggestion
    Tracks an event with the specified name and properties.
```

tests/test_send_telemetry_appinsights.py
```diff
@@ -0,0 +1,102 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
```
zhiyuanliang-ms, line 6: redundant empty line

samples/feature_variant_sample_with_telemetry.py
```diff
@@ -0,0 +1,33 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
```
zhiyuanliang-ms, line 6: redundant empty line

samples/feature_variant_sample_with_telemetry.py
```diff
@@ -0,0 +1,33 @@
...
+import json
+import os
+import sys
+from random_filter import RandomFilter
+from featuremanagement import FeatureManager
+from featuremanagement.opentelemetry import publish_telemetry
+
```
zhiyuanliang-ms, line 13: Need one more empty line to make it consistent with other files

featuremanagement/opentelemetry/_send_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+EVENT_NAME = "FeatureEvaluation"
+
+_event_logger.propagate = False
+
+
+class _FeatureManagementEventsExtension:
+    _initialized = False
+
+    @staticmethod
+    def initialize():
```
zhiyuanliang-ms, line 42: I think the name `ensureInit` would be more appropriate.

featuremanagement/opentelemetry/_send_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    if not HAS_OPENTELEMETRY_SDK:
+        return
+    if event_properties is None:
+        event_properties = {}
+    if user:
+        event_properties[TARGETING_ID] = user
+    if HAS_AZURE_MONITOR_EVENTS_EXTENSION:
+        azure_monitor_track_event(event_name, event_properties)
+        return
```
zhiyuanliang-ms, line 68: I don't understand why if HAS_AZURE_MONITOR_EVENTS_EXTENSION, _event_logger will be skipped?

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -0,0 +1,62 @@
...
+    event[FEATURE_NAME] = evaluation_event.feature.name
+    event[ENABLED] = str(evaluation_event.enabled)
+    if evaluation_event.user:
+        event[TARGETING_ID] = evaluation_event.user
+
+    if evaluation_event.reason and evaluation_event.reason != VariantAssignmentReason.NONE:
+        event[VARIANT] = evaluation_event.variant.name
+        event[REASON] = evaluation_event.reason.value
+
+    event["ETag"] = evaluation_event.feature.telemetry.metadata.get("etag", "")
```
zhiyuanliang-ms: We should publish all fields in feature.telemetry.metadata like [this](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L410)

zhiyuanliang-ms: If python provider sets "etag", "feature_flag_reference", "feature_flag_id" fields to the telemetry metadata, I think we should have another PR to update them to pascal case names.

samples/feature_variant_sample_with_telemetry.py
zhiyuanliang-ms, line 6: redundant empty line

samples/feature_variant_sample_with_telemetry.py
```diff
@@ -0,0 +1,33 @@
...
+with open(script_directory + "/formatted_feature_flags.json", "r", encoding="utf-8") as f:
+    feature_flags = json.load(f)
+
+# Initialize the feature manager with telemetry callback
+feature_manager = FeatureManager(
+    feature_flags, feature_filters=[RandomFilter()], on_feature_evaluated=publish_telemetry
+)
+
+# Evaluate the feature flag for the user
+print(feature_manager.get_variant("TestVariants", "Adam").configuration)
```
rossgrambo, line 33: Could we add a call to track_event after this to demonstrate the non-impression event part as well?

featuremanagement/_featuremanager.py
```diff
@@ -90,54 +93,64 @@ def __init__(self, configuration, **kwargs):
...
-            feature_flag.variants, feature_flag.allocation.default_when_enabled, True
+        evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+        if not evaluation_event.feature.allocation:
+            evaluation_event.enabled = True
+            return
+        FeatureManager._check_variant_override(
+            evaluation_event.feature.variants,
+            evaluation_event.feature.allocation.default_when_enabled,
+            True,
+            evaluation_event,
```
zhiyuanliang-ms, line 130:
```suggestion
            evaluation_event
```

featuremanagement/aio/_featuremanager.py
```diff
@@ -43,54 +47,63 @@ def __init__(self, configuration, **kwargs):
...
-        return FeatureManager._check_variant_override(
-            feature_flag.variants, feature_flag.allocation.default_when_enabled, True
+        if not evaluation_event.feature.allocation:
+            evaluation_event.enabled = True
+            return
+        FeatureManager._check_variant_override(
+            evaluation_event.feature.variants,
+            evaluation_event.feature.allocation.default_when_enabled,
+            True,
+            evaluation_event,
```
zhiyuanliang-ms, line 82:
```suggestion
            evaluation_event
```

featuremanagement/aio/_featuremanager.py
```diff
@@ -43,54 +47,63 @@ def __init__(self, configuration, **kwargs):
...
-        return FeatureManager._check_variant_override(
-            feature_flag.variants, feature_flag.allocation.default_when_disabled, False
+        if not evaluation_event.feature.allocation:
+            evaluation_event.enabled = False
+            return
+        FeatureManager._check_variant_override(
+            evaluation_event.feature.variants,
+            evaluation_event.feature.allocation.default_when_disabled,
+            False,
+            evaluation_event,
```
zhiyuanliang-ms, line 64:
```suggestion
            evaluation_event
```

featuremanagement/aio/_featuremanager.py
```diff
@@ -43,54 +47,63 @@ def __init__(self, configuration, **kwargs):
...
 
-        :param FeatureFlag feature_flag: Feature flag object.
-        :return: EvaluationEvent
+        :param EvaluationEvent evaluation_event: Evaluation event object.
         """
-        if not feature_flag.allocation:
-            return EvaluationEvent(enabled=True)
-        return FeatureManager._check_variant_override(
-            feature_flag.variants, feature_flag.allocation.default_when_enabled, True
+        if not evaluation_event.feature.allocation:
```
zhiyuanliang-ms, line 75: Forget to set evaluation_event.reason here.

featuremanagement/aio/_featuremanager.py
```diff
@@ -43,54 +47,63 @@ def __init__(self, configuration, **kwargs):
...
 
-        :param FeatureFlag feature_flag: Feature flag object.
-        :return: EvaluationEvent
+        :param EvaluationEvent evaluation_event: Evaluation event object.
         """
-        if not feature_flag.allocation:
-            return EvaluationEvent(enabled=False)
-        return FeatureManager._check_variant_override(
-            feature_flag.variants, feature_flag.allocation.default_when_disabled, False
+        if not evaluation_event.feature.allocation:
```
zhiyuanliang-ms, line 57: Forget to set evaluation_event.reason here.

featuremanagement/_featuremanager.py
```diff
@@ -283,31 +315,24 @@ def _check_feature_filters(self, feature_flag, targeting_context, **kwargs):
...
+            if not feature_flag.allocation:
+                if evaluation_event.enabled:
+                    evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+                    return
+                evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED
+                return
+            if not evaluation_event.enabled:
+                FeatureManager._check_default_disabled_variant(evaluation_event)
+                evaluation_event.variant = self._variant_name_to_variant(
+                    feature_flag, feature_flag.allocation.default_when_enabled
```
zhiyuanliang-ms, line 331: default_when_disabled

featuremanagement/aio/_featuremanager.py
```diff
@@ -224,40 +248,40 @@ async def _check_feature_filters(self, feature_flag, targeting_context, **kwargs
...
+            if not feature_flag.allocation:
+                if evaluation_event.enabled:
+                    evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+                    return
+                evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED
+                return
+            if not evaluation_event.enabled:
+                FeatureManager._check_default_disabled_variant(evaluation_event)
+                evaluation_event.variant = self._variant_name_to_variant(
+                    feature_flag, feature_flag.allocation.default_when_enabled
```
zhiyuanliang-ms, line 272: defautl_when_disabled

featuremanagement/aio/_featuremanager.py
```diff
@@ -224,40 +248,40 @@ async def _check_feature_filters(self, feature_flag, targeting_context, **kwargs
...
-            if feature_flag.allocation:
-                variant_name = feature_flag.allocation.default_when_enabled
-        else:
-            evaluation_event = FeatureManager._check_default_disabled_variant(feature_flag)
-            if feature_flag.allocation:
-                variant_name = feature_flag.allocation.default_when_disabled
-        evaluation_event.variant = self._variant_name_to_variant(feature_flag, variant_name)
-        evaluation_event.feature = feature_flag
-        return evaluation_event
+            variant_name = self._assign_variant(targeting_context, evaluation_event)
```
zhiyuanliang-ms, line 276: Inconsistent with another feature manager. I am ok with another version.

featuremanagement/_featuremanager.py
```diff
@@ -327,23 +352,25 @@ def _check_feature(self, feature_flag_id, targeting_context, **kwargs):
...
         if not feature_flag:
             logging.warning("Feature flag %s not found", feature_flag_id)
             # Unknown feature flags are disabled by default
-            return EvaluationEvent(enabled=False)
+            return evaluation_event
 
         if not feature_flag.enabled:
             # Feature flags that are disabled are always disabled
-            evaluation_event = FeatureManager._check_default_disabled_variant(feature_flag)
+            FeatureManager._check_default_disabled_variant(evaluation_event)
```
zhiyuanliang-ms, line 363: check_variant_override is called in _check_default_disabled_variant. However, status_override should not work when feature_flag.enabled == false. Current behavior is wrong.

Currently, you have `check_variant_override` called in `_check_default_disabled_variant`, `_check_default_enabled_variant` and `_assign_variant`. The code path is complex.

I suggest we have the code path as below

``` python
def _check_feature():
    feature_flag = _get_feature_flag()
    ...

    evaluation_event = EvaluationEvent(feature_flag)

    _check_feature_filters(evaluation_event)

    _assign_allocation(evaluation_event)

    if feature_flag.enabled:
        _check_variant_override()
```

`_assign_allocation` follows
https://github.com/microsoft/FeatureManagement-Python/pull/24#issuecomment-2219830895

I am ok with the current change on the `_assign_variant` method where `default_when_enabled` is also handled there.

zhiyuanliang-ms (changes requested):
check_variant_override is called in _check_default_disabled_variant. However, status_override should not work when feature_flag.enabled == false. Current behavior is wrong.

Currently, you have `check_variant_override` called in `_check_default_disabled_variant`, `_check_default_enabled_variant` and `_assign_variant`. The code path is complex.

I suggest we have the code path as below

``` python
def _check_feature():
    feature_flag = _get_feature_flag()
    ...

    evaluation_event = EvaluationEvent(feature_flag)

    _check_feature_filters(evaluation_event)

    _assign_allocation(evaluation_event)

    if feature_flag.enabled:
        _check_variant_override()
```

`_assign_allocation` follows
https://github.com/microsoft/FeatureManagement-Python/pull/24#issuecomment-2219830895

I am ok with the current change on the `_assign_variant` method where `default_when_enabled` is also handled there.

featuremanagement/_featuremanager.py
mrm9084: @rossgrambo how does the DotNet Feature Management library work for this?

featuremanagement/opentelemetry/_send_telemetry.py
mrm9084: It makes sure that the logging is handled by the LoggingHandler on line 47, and not any other. This makes sure open telemetry gets our log. App Insights does the same thing.

featuremanagement/opentelemetry/_send_telemetry.py
mrm9084: We are making our own copy of this, https://github.com/microsoft/ApplicationInsights-Python/blob/main/azure-monitor-events-extension/azure/monitor/events/extension/_events.py, which follows this pattern.

featuremanagement/_featuremanager.py
rossgrambo: When dotnet had the callback- we did not call the callback, but it was specifically a "TelemetryPublisher" that we were not calling. 

Today, if telemetry is disabled- dotnet won't create an activity or event- so there is no non-telemetry callback behavior. But- the reason why is because we didn't want to add too much to their telemetry, which would happen if OpenTelemetry ingested it.

So either:
1. To be in pairity with dotnet- don't call the callback at all
2. Commit to the idea that this callback doesn't have to be telemetry related- and do the check only on the telemetry specific things.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: @rossgrambo has this changed? When I wrote this originally, I copied the say outputs as DotNet.

featuremanagement/_featuremanager.py
mrm9084: This formatting is correct for the Black formatter. And the format we follow for both the provider and feature management libraries.

featuremanagement/aio/_featuremanager.py
mrm9084: I should have fixed all of these. Seems like only the sync file got the review comments.

featuremanagement/aio/_featuremanager.py
mrm9084: This should be fixed. I went through an made sure the sync and async code matches. After I finish this PR I really need to make a base class both sync and async extend that should remove a good amount of duplicate code.

featuremanagement/_featuremanager.py
mrm9084: @rossgrambo, I think we should default to parity and if DotNet changes we can too.

featuremanagement/azuremonitor/_send_telemetry.py
zhiyuanliang-ms: .net provider: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/preview/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs#L44

featuremanagement/aio/_featuremanager.py
zhiyuanliang-ms: > After I finish this PR I really need to make a base class both sync and async extend that should remove a good amount of duplicate code.

YESSSS!

featuremanagement/_featuremanager.py
rossgrambo: > @rossgrambo, I think we should default to parity and if DotNet changes we can too.

Then I think we should rename the callback to be "telemetry_callback" or something to make it clear that `telemetry.enabled = false` means it won't be called.

Personally- I prefer breaking from .net in regards to this callback- and placing the check inside of `publish_telemetry`.

featuremanagement/azuremonitor/_send_telemetry.py
rossgrambo: > @rossgrambo has this changed? When I wrote this originally, I copied the say outputs as DotNet.

You might be thinking of the provider. `FeatureManagement` never knew what an ETag was- it just grabs everything in metadata and blindly adds it.

mrm9084: This should be correct now.

featuremanagement/_featuremanager.py
mrm9084: @rossgrambo, @zhiyuanliang-ms, doesn't the fact that this override doesn't work mean that it is more complex as in the `is_enabled` case we need to return false always if the feature flag isn't enabled, but the correct override variant in the case of `get_variant`?

featuremanagement/aio/_featuremanager.py
mrm9084: Resolving this we have a larger conversation bellow. https://github.com/microsoft/FeatureManagement-Python/pull/24#discussion_r1673464625

featuremanagement/_featuremanager.py
rossgrambo: > However, status_override should not work when feature_flag.enabled == false. Current behavior is wrong.

I don't think this is correct. We run both IsEnabled and Variant allocation on both .IsEnabled and .GetVariant calls, regardless of if the flag is enabled or disabled- then apply the override https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L282

rossgrambo: But according to the flowcharts it should be disabled if the flag is disabled... So actually you're right.

rossgrambo: > doesn't the fact that this override doesn't work mean that it is more complex as in the is_enabled case we need to return false always if the feature flag isn't enabled, but the correct override variant in the case of get_variant?

Yes but it's not complex if you just do both everytime. Dotnet just always determines both the IsEnabled result and the Variant result (which will be defaultWhenDisabled in the disabled case). Then if IsEnabled was the calling function- just returns the bool, and if GetVariant was the calling function, returns the variant.

mrm9084: This code works the same as DotNet, which is why is does the slightly more complex operation, but it does simplify the whole logic as variants and is_enabled mostly work the same.

featuremanagement/aio/_featuremanager.py
```diff
@@ -43,54 +47,65 @@ def __init__(self, configuration, **kwargs):
...
-        :return: EvaluationEvent
+        :param EvaluationEvent evaluation_event: Evaluation event object.
         """
-        if not feature_flag.allocation:
-            return EvaluationEvent(enabled=False)
-        return FeatureManager._check_variant_override(
-            feature_flag.variants, feature_flag.allocation.default_when_disabled, False
+        evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED
+        if not evaluation_event.feature.allocation:
+            evaluation_event.enabled = False
```
zhiyuanliang-ms, line 59: In line 286,
``` python
if not evaluation_event.enabled:
    FeatureManager._check_default_disabled_variant(evaluation_event)
```

Setting `evaluation_event.enabled = False` in `_check_default_disabled_variant` is redundant.

In line 327,

``` python
if not feature_flag.enabled:
    # Feature flags that are disabled are always disabled
    FeatureManager._check_default_disabled_variant(evaluation_event)
    if feature_flag.allocation:
        variant_name = feature_flag.allocation.default_when_disabled
        evaluation_event.variant = self._variant_name_to_variant(feature_flag, variant_name)
    evaluation_event.feature = feature_flag

    # If a feature flag is disabled and override can't enable it
    evaluation_event.enabled = False
    return evaluation_event
```

After calling  `_check_default_disabled_variant`, you set `evaluation_event.enabled = False` again.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -0,0 +1,64 @@
...
+    event = {}
+    event[FEATURE_NAME] = evaluation_event.feature.name
+    event[ENABLED] = str(evaluation_event.enabled)
+
+    if evaluation_event.reason and evaluation_event.reason != VariantAssignmentReason.NONE:
+        event[VARIANT] = evaluation_event.variant.name
+        event[REASON] = evaluation_event.reason.value
+
+    for metadata_key, metadata_value in evaluation_event.feature.telemetry.metadata.items():
+        if metadata_key not in event:
```
zhiyuanliang-ms, line 62: .NET Provider will put these fields in feature flag telemetry metadata.
``` C#
public const string ETag = "ETag";
public const string FeatureFlagId = "FeatureFlagId";
public const string FeatureFlagReference = "FeatureFlagReference";
```
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/preview/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementConstants.cs#L44

However,in this [PR](https://github.com/Azure/azure-sdk-for-python/pull/35254/files#diff-a38814ad69b5c0a6d82fe7946285bb3408209336deb2ba83d8b8808d13ea4657R20)

python provider will put these fields
``` python
ETAG_KEY = "etag"
FEATURE_FLAG_REFERENCE_KEY = "feature_flag_reference"
FEATURE_FLAG_ID_KEY = "feature_flag_id"
```

featuremanagement/azuremonitor/_send_telemetry.py
zhiyuanliang-ms: The telemetry metadata field name is inconsistent with .NET provider.

featuremanagement/aio/_featuremanager.py
mrm9084: For the second part, I still want to set the variant info for is_enabled as it will be provided to telemetry either way. And I believe that is how DotNet also provides things.

featuremanagement/_featuremanager.py
```diff
@@ -90,54 +93,64 @@ def __init__(self, configuration, **kwargs):
...
-        :return: EvaluationEvent
+        :param EvaluationEvent evaluation_event: Evaluation event object.
         """
-        if not feature_flag.allocation:
-            return EvaluationEvent(enabled=True)
-        return FeatureManager._check_variant_override(
-            feature_flag.variants, feature_flag.allocation.default_when_enabled, True
+        evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+        if not evaluation_event.feature.allocation:
+            evaluation_event.enabled = True
```
zhiyuanliang-ms, line 123: Could you also remove this please?

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: Will be updated in the Provider PR once Auto Failover has been completed.

zhiyuanliang-ms (approved):

PR #26, opened 2024-05-13
Targeting Update

mrm9084:
# Description

Moving from telemetry PR. change that switches to targeting_context


featuremanagement/_featuremanager.py
```diff
@@ -83,30 +81,32 @@ def __init__(self, configuration, **kwargs):
     @staticmethod
     def _check_default_disabled_variant(feature_flag):
         if not feature_flag.allocation:
-            return False
-        return FeatureManager._check_variant_override(
+            return EvaluationEvent(enabled=False)
+        evaluation_event = FeatureManager._check_variant_override(
```
rossgrambo, line 85: Could just return here. Also in _check_default_enabled_variant

featuremanagement/_featuremanager.py
```diff
@@ -116,32 +114,28 @@ def _is_targeted(context_id):
 
         return (context_marker / (2**32 - 1)) * 100
 
-    def _assign_variant(self, feature_flag, **kwargs):
+    def _assign_variant(self, feature_flag, targeting_context):
```
rossgrambo, line 117: I'd expect this function to set `evaluation_event.variant` and return the `evaluation_event`, rather than a double return. 

Looking at it's usage- I'd actually expect evaluation_event to be passed in as a param and modified directly.

featuremanagement/_featuremanager.py
```diff
@@ -154,107 +148,155 @@ def _variant_name_to_variant(self, feature_flag, variant_name):
...
-        :param str feature_flag_id: Name of the feature flag
-        :paramtype feature_flag_id: str
-        :return: Name of the variant
-        :rtype: str
+        :param str feature_flag_id: Name of the feature flag.
+        :return: True if the feature flag is enabled for the given context.
+        :rtype: bool
         """
-        return self._check_feature(feature_flag_id, **kwargs)["variant"]
+        targeting_context = TargetingContext()
```
rossgrambo, line 170: This is a nice way to offer both. Could pull this out into a helper since it's used in both is_enabled and get_variant

featuremanagement/aio/_featuremanager.py
```diff
@@ -120,106 +116,115 @@ def _variant_name_to_variant(self, feature_flag, variant_name):
...
-                return result
+                evaluation_event.variant = self._variant_name_to_variant(feature_flag, variant_name)
+                evaluation_event.feature = feature_flag
+                return evaluation_event
 
         variant_name = None
-        if result["enabled"]:
-            result["enabled"] = FeatureManager._check_default_enabled_variant(feature_flag)
+        if evaluation_event.enabled:
+            evaluation_event = FeatureManager._check_default_enabled_variant(feature_flag)
```
rossgrambo, line 179: There's a lot of evaluation event re-creation happening. I would expect a single evaluation event to be created at the start and for that evaluation event to be the one returned. I find it hard to follow the code in the current state. Some comments could help as well

featuremanagement/_featuremanager.py
mrm9084: So, I updated part of this. The `evaluation_event` has been removed. As for the part before that. This method just find the name of the variant. It doesn't get attached to the `evaluation_event` until after we return and check that we returned a none `None` value. See line 241.

featuremanagement/aio/_featuremanager.py
mrm9084: So, I think I cleared the first part up a bit. `EvaluationEvent` is now only created once we need it. I'll try adding some more comments.

rossgrambo (approved):

PR #27, opened 2024-05-22
Release Prep

mrm9084:
# Description

* Adds Changelog with date
* Fixes Targeting Context Sphinx doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 New Feature Management library.
 
-Provides the ability to manage features in a project. Enables:
+Provides the ability to manage feature flags in a project. Enables:
 
 * Loading of feature flags from a file, see: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json
 * Loading of feature flags from Azure App Configuration.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
 # Release History
 
-## 1.0.0b1 (Unreleased)
+## 1.0.0b1 (05/22/2024)
+
+New Feature Management library.
+
+Provides the ability to manage features in a project. Enables:
+
+* Loading of feature flags from a file, see: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json
+* Loading of feature flags from Azure App Configuration.
+* Checking if a feature is enabled.
+* Default feature filters: TimeWindowFilter, TargetingFilter.
+* Custom feature filters.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
 # Release History
 
-## 1.0.0b1 (Unreleased)
+## 1.0.0b1 (05/22/2024)
+
+New Feature Management library.
+
+Provides the ability to manage features in a project. Enables:
```
albertofori, line 7: ...to manage feature **flags**?

rossgrambo (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+
+* Loading of feature flags from a file, see: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json
```
albertofori, line 9: This definition references a feature flag definition that defines variants. Is that intended?

featuremanagement/_models/_targeting_context.py
```diff
@@ -14,6 +14,17 @@ class TargetingContext(NamedTuple):
 
     # The user ID
     user_id: str = ""
+    """
```
albertofori, line 17: Not sure but should this doc string rather be on top? Also, I see we already have the same description above on line 15. Should we just move this docstring there,

CHANGELOG.md
mrm9084: This is the only one that has the prefix "feature_management.feature_flags".

featuremanagement/_models/_targeting_context.py
mrm9084: This doc string needs to be here for sphinx to pick it up. I already tested it.

mrm9084: https://microsoft.github.io/FeatureManagement-Python/html/index.html#featuremanagement.TargetingContext

CHANGELOG.md
albertofori: Got it, and that is the format that is required in this version of the feature management library, I suppose?

featuremanagement/_models/_targeting_context.py
albertofori: Then perhaps, we can remove the comment on top if that is the case.

CHANGELOG.md
mrm9084: Yes, it's the only one we support for this and in JS.

featuremanagement/_models/_targeting_context.py
mrm9084: Refresh the PR, I already pushed the change.

CHANGELOG.md
albertofori: Sounds good!

albertofori (approved):

Issue #33, opened 2024-07-10
Add type hinting


zhiyuanliang-ms:
Type hinting is supported for python 3.5+. It will be very helpful for Improving code readability and maintainability.

https://docs.python.org/3/library/typing.html

mrm9084:
With this https://github.com/microsoft/FeatureManagement-Python/pull/36, I think we can close this PR. mypy has been added to the PR GitHub action with strict settings on.

PR #35, opened 2024-07-31
Static Type Checking Intro

mrm9084:
# Description

This PR starts to add static type checking to the Feature Management library.
See: mypy.ini for details on what is enabled and what isn't.

Starts work on this: https://github.com/microsoft/FeatureManagement-Python/issues/33

Note: Currently this will not be part of the build step until more is done.
Note: Due to some of the type check added to pass mypy some code changes were required.

# Contribution checklist:

### DIFF
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -4,17 +4,7 @@
 # license information.
 # -------------------------------------------------------------------------
 from abc import ABC, abstractmethod
-from typing import Mapping, Callable, TypedDict, Any
-from typing_extensions import Unpack
-
-
-class RequestParams(TypedDict):
-    """
-    Type for request parameters.
-    """
-
-    key: str
-    value: Any
+from typing import Mapping, Callable, Dict, Any
 
 
 class FeatureFilter(ABC):
@@ -23,7 +13,7 @@ class FeatureFilter(ABC):
     """
 
     @abstractmethod
-    def evaluate(self, context: Mapping, **kwargs: Unpack[RequestParams]) -> bool:
+    def evaluate(self, context: Mapping, **kwargs: Dict[str, Any]) -> bool:
         """
         Determine if the feature flag is enabled for the given context.
 
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -4,6 +4,17 @@
 # license information.
 # -------------------------------------------------------------------------
 from abc import ABC, abstractmethod
+from typing import Mapping, Callable, TypedDict, Any
+from typing_extensions import Unpack
+
+
+class RequestParams(TypedDict):
+    """
+    Type for request parameters.
+    """
+
+    key: str
+    value: Any
 
 
 class FeatureFilter(ABC):
@@ -12,7 +23,7 @@ class FeatureFilter(ABC):
     """
 
     @abstractmethod
-    def evaluate(self, context, **kwargs):
+    def evaluate(self, context: Mapping, **kwargs: Unpack[RequestParams]) -> bool:
         """
         Determine if the feature flag is enabled for the given context.
 
@@ -20,7 +31,7 @@ class FeatureFilter(ABC):
         """
 
     @property
-    def name(self):
+    def name(self) -> str:
         """
         Get the name of the filter.
 
@@ -32,7 +43,7 @@ class FeatureFilter(ABC):
         return self.__class__.__name__
 
     @staticmethod
-    def alias(alias):
+    def alias(alias: str) -> Callable:
         """
         Decorator to set the alias for the filter.
 
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -39,7 +39,7 @@ class FeatureFilter(ABC):
         :rtype: str
         """
         if hasattr(self, "_alias"):
-            return self._alias # type: ignore
+            return self._alias  # type: ignore
         return self.__class__.__name__
 
     @staticmethod
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -52,7 +52,7 @@ class FeatureFilter(ABC):
         :rtype: callable
         """
 
-        def wrapper(cls):
+        def wrapper(cls) -> Callable:  # type: ignore
             cls._alias = alias  # pylint: disable=protected-access
             return cls
 
--- a/featuremanagement/_featurefilters.py
+++ b/featuremanagement/_featurefilters.py
@@ -39,7 +39,7 @@ class FeatureFilter(ABC):
         :rtype: str
         """
         if hasattr(self, "_alias"):
-            return self._alias
+            return self._alias # type: ignore
         return self.__class__.__name__
 
     @staticmethod
@@ -49,10 +49,10 @@ class FeatureFilter(ABC):
 
         :param str alias: Alias for the filter.
         :return: Decorator.
-        :rtype: callable
+        :rtype: Callable
         """
 
-        def wrapper(cls) -> Callable:  # type: ignore
+        def wrapper(cls) -> Any:  # type: ignore
             cls._alias = alias  # pylint: disable=protected-access
             return cls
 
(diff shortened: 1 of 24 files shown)


featuremanagement/_featurefilters.py
```diff
@@ -12,36 +23,36 @@ class FeatureFilter(ABC):
...
+    def name(self) -> str:
         """
         Get the name of the filter.
 
         :return: Name of the filter, or alias if it exists.
         :rtype: str
         """
         if hasattr(self, "_alias"):
-            return self._alias
+            return self._alias  # type: ignore
```
zhiyuanliang-ms, line 42: I don't understand the comment "type: ignore" here.

featuremanagement/_featuremanager.py
```diff
@@ -37,8 +37,8 @@ def __init__(self, configuration, **kwargs):
                 raise ValueError("Custom filter must be a subclass of FeatureFilter")
             self._filters[feature_filter.name] = feature_filter
 
-    @overload
-    def is_enabled(self, feature_flag_id, user_id, **kwargs):
+    @overload  # type: ignore
+    def is_enabled(self, feature_flag_id: str, user_id: str, **kwargs: Dict) -> bool:
```
zhiyuanliang-ms, line 41: May I know the reason why here you have `**kwargs: Dict`, but for `FeatureFilter.Evaluate` you have `**kwargs: Unpack[RequestParams]`?

featuremanagement/_featurefilters.py
```diff
@@ -4,6 +4,17 @@
 # license information.
 # -------------------------------------------------------------------------
 from abc import ABC, abstractmethod
+from typing import Mapping, Callable, TypedDict, Any
+from typing_extensions import Unpack
+
+
+class RequestParams(TypedDict):
```
zhiyuanliang-ms, line 11: This is just one key-value pair. So should it be `RequestParam` (no 's')?

featuremanagement/_featurefilters.py
```diff
@@ -12,36 +23,36 @@ class FeatureFilter(ABC):
     """
 
     @abstractmethod
-    def evaluate(self, context, **kwargs):
+    def evaluate(self, context: Mapping, **kwargs: Unpack[RequestParams]) -> bool:
```
zhiyuanliang-ms, line 26: I don't know whether my current understanding is correct. If I say something stupid please forgive me. 

IMO, `**kwargs` itself is a `dict` already. For `foo(**kwargs)`, you can do something like: `for key, value in kwargs.items(): # do something#`

But now, you have `foo(**kwargs: Unpack[RequestParams])`. The way you call `foo` will be like: `foo(param1={"key": "A", "value": 1}, param2={"key": "B", "value": 2})`.

It will be `foo(param1={"key": "A", "value": 1}` vs  `foo(A=1)`

I think `RequestParams` class is over-engineered.

featuremanagement/_featuremanager.py
```diff
@@ -87,13 +92,22 @@ def get_variant(self, feature_flag_id, *args, **kwargs):
...
+            and callable(self._on_feature_evaluated)
+        ):
             result.user = targeting_context.user_id
             self._on_feature_evaluated(result)
         return result.variant
 
-    def _check_feature_filters(self, evaluation_event, targeting_context, **kwargs):
+    def _check_feature_filters(
+        self, evaluation_event: EvaluationEvent, targeting_context: TargetingContext, **kwargs: Dict
+    ) -> None:
```
zhiyuanliang-ms, line 107: Is this new line intended?

featuremanagement/_featurefilters.py
mrm9084: This comment and those like it are things that the type hints for Mypy are not correct yet. This one in particular I haven't figured out how to get it to work correctly. So right now I'm suppressing the error with this.

featuremanagement/_featuremanager.py
mrm9084: I'm actually unsure which one it should be. I imagine it's should be a `Dict[str, Any]` but I couldn't find any good documentation on it.

featuremanagement/_featurefilters.py
mrm9084: I agree, I was looking to figure out what the correct value of it should be and the `RequestParams` solution is one of them. the other being a `Dict[str, Any]`. I'm good with either one, it seems like you lean towards the `Dict[str, Any]` option.

featuremanagement/_featuremanager.py
mrm9084: Yes, the line with the type hints is now over 120 characters, so our formatter requires it to be on a new line.

featuremanagement/_featuremanager.py
```diff
@@ -28,17 +28,17 @@ class FeatureManager(FeatureManagerBase):
     evaluated.
     """
 
-    def __init__(self, configuration, **kwargs):
+    def __init__(self, configuration: Mapping, **kwargs: Dict[str, Any]):
         super().__init__(configuration, **kwargs)
-        filters = [TimeWindowFilter(), TargetingFilter()] + kwargs.pop(PROVIDED_FEATURE_FILTERS, [])
+        filters = [TimeWindowFilter(), TargetingFilter()] + cast(List, kwargs.pop(PROVIDED_FEATURE_FILTERS, []))
```
rossgrambo, line 33: Could we do Sequence instead of List since we probably don't plan to edit it?

featuremanagement/_featuremanager.py
```diff
@@ -75,7 +80,9 @@ def get_variant(self, feature_flag_id, user_id, **kwargs):
         :rtype: Variant
         """
 
-    def get_variant(self, feature_flag_id, *args, **kwargs):
+    def get_variant(  # type: ignore
```
rossgrambo, line 83: Why do we need type ignore here?

featuremanagement/_featuremanager.py
mrm9084: Just checked, mypy doesn't like `Sequence` it doesn't allow us to add the two lists together.

featuremanagement/_featuremanager.py
mrm9084: mypy currently throws an error saying `Overloaded function implementation does not accept all possible arguments of signature 1`. I'm trying to limit the size of each change as there is already a lot of them in this PR. If you look at the `mypy.ini` file I have a couple things set to false as they would require even more changes.

Also, I think this specific one is about the issue we have talked about before. I'll get to this in another PR.

featuremanagement/_models/_feature_flag.py
```diff
@@ -126,15 +126,15 @@ def telemetry(self):
...
+    def _validate(self) -> None:
         if not isinstance(self._id, str):
             raise ValueError(f"Invalid setting 'id' with value '{self._id}' for feature '{self._id}'.")
         if not isinstance(self._enabled, bool):
             raise ValueError(f"Invalid setting 'enabled' with value '{self._enabled}' for feature '{self._id}'.")
         self.conditions._validate(self._id)  # pylint: disable=protected-access
 
 
-def _convert_boolean_value(enabled):
+def _convert_boolean_value(enabled: Union[str, bool], feature_name: str) -> bool:
```
rossgrambo, line 137: Feels weird to pass in the name here only for the error message.

featuremanagement/_models/_variant.py
```diff
@@ -13,22 +14,22 @@ class Variant:
...
+    def name(self) -> str:
         """
         The name of the variant.
         :rtype: str
         """
         return self._name
 
     @property
-    def configuration(self):
+    def configuration(self) -> Union[Any, Dict]:
```
rossgrambo, line 30: Does a Union here provide any benefit- since one of the types is already Any?

rossgrambo (approved):

featuremanagement/_models/_feature_flag.py
mrm9084: This is so we can match the error message as we have in the other libraries. It also feels weird to do the check outside of this.

featuremanagement/_models/_variant.py
mrm9084: This is the variant object, so I was sort of thinking it made sense sort of as it could be a config or a json. But i'm good making it just Any.

rossgrambo (approved):

PR #41, opened 2024-09-11
New telemetry values

mrm9084:
# Description

Adds VariantAllocationPercentage, DefaultWhenEnabled, and AllocationId to telemetry.

# Contribution checklist:
- [x] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [x] **I have read the [contribution guidelines](https://github.com/microsoft/FeatureManagement-Python/blob/main/CONTRIBUTING.md).**
- [ ] Pull request includes test coverage for the included changes.

- [ ] Title of the pull request is clear and informative.
- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.


featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,54 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
+                event["VariantAssignmentPercentage"] = str(allocation_percentage)
+
+    # DefaultWhenEnabled
+    if evaluation_event.feature.allocation and evaluation_event.feature.allocation.default_when_enabled:
+        event["DefaultWhenEnabled"] = evaluation_event.feature.allocation.default_when_enabled
+
+    # AllocationId
+    allocation_id = evaluation_event.feature.telemetry.metadata.get("AllocationId")
+    if allocation_id:
+        event["AllocationId"] = allocation_id
```
rossgrambo, line 101: This is handled automatically by the metadata below

```suggestion
```

rossgrambo (approved):

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,49 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
+                    ):
+                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
+
+            event["VariantAssignmentPercentage"] = str(100 - allocation_percentage)
+        elif evaluation_event.reason == VariantAssignmentReason.PERCENTILE:
+            if evaluation_event.feature.allocation and evaluation_event.feature.allocation.percentile:
+                allocation_percentage = 0
+                for allocation in evaluation_event.feature.allocation.percentile:
+                    if (
+                        evaluation_event.variant
```
zhiyuanliang-ms, line 87: It would be better to check whether evaluation_event.variant is None in advance.

zhiyuanliang-ms: Is it a possible case that variantAssignmentReason is not None but variant is None? I cannot imagine it will happen.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,49 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
-    if evaluation_event.feature and evaluation_event.feature.telemetry:
+        if evaluation_event.reason == VariantAssignmentReason.DEFAULT_WHEN_ENABLED:
+            allocation_percentage = 0
+
+            if evaluation_event.feature.allocation and evaluation_event.feature.allocation.percentile:
+                for allocation in evaluation_event.feature.allocation.percentile:
+                    if (
+                        evaluation_event.variant
+                        and allocation.variant == evaluation_event.variant.name
+                        and allocation.percentile_to
```
zhiyuanliang-ms, line 77: Why percentile_to is checked, but percentile_from is not checked?

zhiyuanliang-ms: I suggest just remove this check since the allocation must be fine, otherwise the feature manager should have failed.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,49 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
+                    ):
+                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
+
+            event["VariantAssignmentPercentage"] = str(100 - allocation_percentage)
```
zhiyuanliang-ms, line 81: Is the reason of this type conversion (number to string) that we want to put it in the customDimension instead of customMetrics of the application insights event?

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,49 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
+                        evaluation_event.variant
+                        and allocation.variant == evaluation_event.variant.name
+                        and allocation.percentile_to
+                    ):
+                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
+                event["VariantAssignmentPercentage"] = str(allocation_percentage)
+
+    # DefaultWhenEnabled
+    if evaluation_event.feature.allocation and evaluation_event.feature.allocation.default_when_enabled:
+        event["DefaultWhenEnabled"] = evaluation_event.feature.allocation.default_when_enabled
```
zhiyuanliang-ms, line 96: Just want to call out that it is possible that variant assignment reason is defaultWhenEnabled but the assigned variant is empty/none. The reason of this design is that if we have empty allocation configuration, (in .NET) we can consider it as:

``` json
{
    "allocation": {
        "default_when_enabled" : ""
    }
}

```

zhiyuanliang-ms: I don't whether it is a better choice to have event["DefaultWhenEnabled"] = "" when there is no allocation.default_when_enabled. But it could be another option for us.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,49 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
+                        evaluation_event.variant
+                        and allocation.variant == evaluation_event.variant.name
+                        and allocation.percentile_to
+                    ):
+                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
+
+            event["VariantAssignmentPercentage"] = str(100 - allocation_percentage)
+        elif evaluation_event.reason == VariantAssignmentReason.PERCENTILE:
+            if evaluation_event.feature.allocation and evaluation_event.feature.allocation.percentile:
+                allocation_percentage = 0
```
zhiyuanliang-ms, line 84: Can you move this line before the above if statement to make it consistent with line 70

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: @zhiyuanliang-ms, I've made some changes already to this in my already existing PR that has some updates. https://github.com/microsoft/FeatureManagement-Python/pull/45

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: I thought I posted this somewhere else. For type reasons we have `from` as an `int` and `to` as an `optional[int]`. We need to verify to as it is optional.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: The type of `event` is `Dict[str, Optional[str]]`

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: @rossgrambo

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: See my latest PR. I already made these changes.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,17 +53,49 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
 
-    if evaluation_event.feature and evaluation_event.feature.telemetry:
+        if evaluation_event.reason == VariantAssignmentReason.DEFAULT_WHEN_ENABLED:
+            allocation_percentage = 0
+
+            if evaluation_event.feature.allocation and evaluation_event.feature.allocation.percentile:
+                for allocation in evaluation_event.feature.allocation.percentile:
+                    if (
+                        evaluation_event.variant
+                        and allocation.variant == evaluation_event.variant.name
```
zhiyuanliang-ms, line 76: @rossgrambo @mrm9084 Can you confirm whether this line is by design please?

rossgrambo: This is incorrect- it's updated in https://github.com/microsoft/FeatureManagement-Python/pull/45/files#diff-e599fecd01f487185b6b19104e39d7a20d310657896ddeee742af944b129945a

PR #45, opened 2024-10-17
Small Fixes

mrm9084:
# Description

* Fixes a bug where no reason would be assigned if your percentile allocation was exactly 100.
* Reworked code a bit to be more performant.
* Added some missing test cases to telemetry

# Contribution checklist:


tests/test_send_telemetry_appinsights.py
```diff
@@ -76,3 +98,81 @@ def test_send_telemetry_appinsights_no_variant(self):
...
+                evaluation_event
+            )
+            mock_track_event.assert_called_once()
+            assert mock_track_event.call_args[0][0] == "FeatureEvaluation"
+            assert mock_track_event.call_args[0][1]["FeatureName"] == "TestFeature"
+            assert mock_track_event.call_args[0][1]["Enabled"] == "True"
+            assert mock_track_event.call_args[0][1]["TargetingId"] == "test_user"
+            assert mock_track_event.call_args[0][1]["Variant"] == "big"
+            assert mock_track_event.call_args[0][1]["VariantAssignmentReason"] == "Percentile"
+            assert mock_track_event.call_args[0][1]["VariantAssignmentPercentage"] == "25"
```
mrm9084, line 177: @rossgrambo this is 25, but if someone not in the 25 assigned percent is assigned the reason will be default when enabled but the variant assignment percentage will be 50.

rossgrambo: Yes that is correct and the weirdest part of what we landed on. 

`VariantAssignmentPercentage` shows the percentage of people expected to get this variant for the same `VariantAssignmentReason`. Most of the time `VariantAssignmentReason` will be percentile. When it's default- then the reason reflects the number of people who will get the default variant.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -53,50 +55,53 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
+        event[REASON] = reason.value
+
+        allocation_percentage = 0
+        if (
+            reason == VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+            and feature.allocation
+            and feature.allocation.percentile
+        ):
+            for allocation in feature.allocation.percentile:
+                if variant and allocation.variant != variant.name and allocation.percentile_to:
```
zhiyuanliang-ms, line 86: I am confused. Should it be allocation.variant == variant.name?

zhiyuanliang-ms: I still feel weird to only check percentile_to here. percentile_from is also used below. I think if we need to validate it, them both of from and to should be validated.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -53,50 +55,53 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
-                allocation_percentage = 0
-                for allocation in evaluation_event.feature.allocation.percentile:
-                    if (
-                        evaluation_event.variant
-                        and allocation.variant == evaluation_event.variant.name
-                        and allocation.percentile_to
-                    ):
+        elif reason == VariantAssignmentReason.PERCENTILE:
+            if feature.allocation and feature.allocation.percentile:
+
```
zhiyuanliang-ms, line 92: Is this empty line intended?

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: For the first part I think I swapped the value in my mind forgetting we subtract later. @rossgrambo, what do you think.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: Don't think so, might have been added while debugging. Removed.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: Also @rossgrambo can you don't check this, as the unit test now breaks with the ==.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -53,50 +55,53 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
-
-            if evaluation_event.feature.allocation and evaluation_event.feature.allocation.percentile:
-                for allocation in evaluation_event.feature.allocation.percentile:
-                    if (
-                        evaluation_event.variant
-                        and allocation.variant == evaluation_event.variant.name
-                        and allocation.percentile_to
-                    ):
-                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
+    if reason and reason != VariantAssignmentReason.NONE:
```
rossgrambo, line 74: I think this check can be removed. The variant should be None already so lines 75-77 will run correctly. The remaining logic has explicit checks that would also be skipped. And we should emit the Reason of None.

featuremanagement/azuremonitor/_send_telemetry.py
rossgrambo: The if statement (just line 86) should be removed. All we want here is the total allocated percentiles. We don't care what variant is allocated in any of them. For someone using the portal- this will always add up to 100.

featuremanagement/azuremonitor/_send_telemetry.py
rossgrambo, line 74: This if should be removed. We plan to always emit the reason- even when it's None. The following lines should work without this if.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -53,50 +55,53 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
-                    if (
-                        evaluation_event.variant
-                        and allocation.variant == evaluation_event.variant.name
-                        and allocation.percentile_to
-                    ):
+        elif reason == VariantAssignmentReason.PERCENTILE:
+            if feature.allocation and feature.allocation.percentile:
+
+                for allocation in feature.allocation.percentile:
+                    if variant and allocation.variant == variant.name and allocation.percentile_to:
```
rossgrambo, line 93: The percentile_to check doesn't make sense to me either. Unless it's checking for None. In which case we should check percentile_from as well.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: Just pushed a fix for this.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: Should there be a default reason if the reason object is `None`?

mrm9084: Did you make this same comment twice, or am I missing something?

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: `percentile_from` can't be `None` but `percentile_to` can. To is type `Optional[int]` from is just in `int`.

mrm9084: Can you please review just the last commit https://github.com/microsoft/FeatureManagement-Python/pull/45/commits/0655a7211c9c0855ff52494698dd19909c0016ed. This is the change that would need to be made to remove the optional.

zhiyuanliang-ms: > percentile_from can't be None but percentile_to can

Interesting. percentile_to should not be Optional[int]. I don't see any reason it could be optional.

rossgrambo: > Can you please review just the last commit https://github.com/microsoft/FeatureManagement-Python/commit/0655a7211c9c0855ff52494698dd19909c0016ed. This is the change that would need to be made to remove the optional.

LGTM

featuremanagement/azuremonitor/_send_telemetry.py
rossgrambo: Yeah- UI bugged out on me and I thought I lost this comment

rossgrambo: I prefer VariantAssignmentReason be an enum with VariantAssignmentReason.NONE as the default value. I don't think it should be optional

mrm9084: @rossgrambo. Sorry I mean `None` as in `null` not `VariantAssignmentReason.NONE`

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: > > percentile_from can't be None but percentile_to can
> 
> Interesting. percentile_to should not be Optional[int]. I don't see any reason it could be optional.

The reason it was, was that I didn't like defaulting to have it being 100, so I put it as optional with a `None` default value. The commit above changed it. We should be good now.

featuremanagement/azuremonitor/_send_telemetry.py
rossgrambo: Yeah that was my understanding. In other languages- an enum can't be null and it defaults to the first value. I see that's not a thing in python- Perhaps at the top of this you can just do:

```python
if (!reason)
    reason = VariantAssignmentReason.NONE

othercode~
```

featuremanagement/azuremonitor/_send_telemetry.py
rossgrambo: Could we make the default 0? Might satisfy both concerns

featuremanagement/_featuremanagerbase.py
```diff
@@ -163,42 +164,46 @@ def _assign_variant(
         variant_name = None
         if not feature or not feature.variants or not feature.allocation:
             return
-        if feature.allocation.user and targeting_context.user_id:
-            for user_allocation in feature.allocation.user:
+
+        allocation = feature.allocation
```
samsadsam, line 168: what's the difference between the variables `feature` and `feature_flag` ?

featuremanagement/_featuremanagerbase.py
```diff
@@ -163,42 +164,46 @@ def _assign_variant(
...
+                percentile_to: int = percentile_allocation.percentile_to
+                if (box == 100 and percentile_to == 100) or (
+                    percentile_allocation.percentile_from <= box < percentile_to
+                ):
                     evaluation_event.reason = VariantAssignmentReason.PERCENTILE
                     variant_name = percentile_allocation.variant
+                    break
+
         if not variant_name:
             FeatureManagerBase._check_default_enabled_variant(evaluation_event)
```
samsadsam, line 197: what does `FeatureManagerBase._check_default_enabled_variant` do exactly?

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -53,50 +55,48 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
-                        and allocation.percentile_to
-                    ):
-                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
-                event["VariantAssignmentPercentage"] = str(allocation_percentage)
+    allocation_percentage = 0
+    if reason == VariantAssignmentReason.DEFAULT_WHEN_ENABLED and feature.allocation and feature.allocation.percentile:
+        for allocation in feature.allocation.percentile:
+            allocation_percentage += allocation.percentile_to - allocation.percentile_from
+
+        event["VariantAssignmentPercentage"] = str(100 - allocation_percentage)
```
mrm9084, line 85: @zhiyuanliang-ms

featuremanagement/_featuremanagerbase.py
```diff
@@ -163,42 +164,46 @@ def _assign_variant(
...
+            context_id = f"{targeting_context.user_id}\n{seed}"
             box: float = self._is_targeted(context_id)
-            for percentile_allocation in feature.allocation.percentile:
-                if box == 100 and percentile_allocation.percentile_to == 100:
-                    variant_name = percentile_allocation.variant
-                if not percentile_allocation.percentile_to:
-                    continue
-                if percentile_allocation.percentile_from <= box < percentile_allocation.percentile_to:
+            for percentile_allocation in allocation.percentile:
+                percentile_to: int = percentile_allocation.percentile_to
```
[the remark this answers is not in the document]
mrm9084: Not after a recent update.

featuremanagement/_featuremanagerbase.py
mrm9084: This if is triggered if no variant was assigned. This is checking what the result is for the default variant that is assigned when the flag is enabled. I think the name is from before a re-work of how evaluation event currently operates.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: We could solve this better by making so the default reason is `VariantAssignmentReason.NONE` and that would make it not `None`able.

featuremanagement/_featuremanager.py
```diff
@@ -143,35 +140,13 @@ def _check_feature(
...
-            return evaluation_event
-
-        if not feature_flag.enabled:
-            # Feature flags that are disabled are always disabled
-            FeatureManager._check_default_disabled_variant(evaluation_event)
-            if feature_flag.allocation:
-                variant_name = feature_flag.allocation.default_when_disabled
-                evaluation_event.variant = self._variant_name_to_variant(feature_flag, variant_name)
-            evaluation_event.feature = feature_flag
+        evaluation_event, done = FeatureManager._check_feature_base(self, feature_flag_id)
```
samsadsam, line 147: Why don't we use the `super()` terminology?

mrm9084: unsure, fixed.

featuremanagement/_featuremanagerbase.py
mrm9084: I think the `feature_flag` value existed prior to the current usage of `evaluation_event`. I've cleaned it up.

featuremanagement/_featuremanagerbase.py
```diff
@@ -225,31 +226,74 @@ def _build_targeting_context(self, args: Tuple[Any]) -> TargetingContext:
...
+            # Unknown feature flags are disabled by default
+            return evaluation_event, True
+
+        if not feature_flag.enabled:
+            # Feature flags that are disabled are always disabled
+            self._check_default_disabled_variant(evaluation_event)
+            if feature_flag.allocation:
+                variant_name = feature_flag.allocation.default_when_disabled
+                evaluation_event.variant = self._variant_name_to_variant(feature_flag, variant_name)
+            evaluation_event.feature = feature_flag
```
samsadsam, line 291: is this assignment necessary given how `evaluation_event` is created?

featuremanagement/_featuremanagerbase.py
```diff
@@ -225,31 +226,74 @@ def _build_targeting_context(self, args: Tuple[Any]) -> TargetingContext:
...
+
+            # If a feature flag is disabled and override can't enable it
+            evaluation_event.enabled = False
```
samsadsam, line 294: I want to confirm that I understand this line.

This line exists for the case when the `default_when_disabled` variant has a `status_override=false`, since a feature flag will always be disabled when the "root" enabled flag is set to `false`. We don't want the disabled state of the feature flag to be overwritten.

Is my understanding correct?

featuremanagement/aio/_featuremanager.py
```diff
@@ -63,7 +60,12 @@ async def is_enabled(self, feature_flag_id: str, *args: Any, **kwargs: Any) -> b
         targeting_context = self._build_targeting_context(args)
 
         result = await self._check_feature(feature_flag_id, targeting_context, **kwargs)
-        if self._on_feature_evaluated and result.feature and result.feature.telemetry.enabled:
+        if (
+            self._on_feature_evaluated
+            and result.feature
+            and result.feature.telemetry.enabled
+            and callable(self._on_feature_evaluated)
```
samsadsam, line 67: I guess the code block within this if statement should be:

```
result.user = targeting_context.user_id
if inspect.iscoroutinefunction(self._on_feature_evaluated):
    await self._on_feature_evaluated(result)
else:
    self._on_feature_evaluated(result)
```

featuremanagement/aio/_featuremanager.py
```diff
@@ -94,7 +96,12 @@ async def get_variant(self, feature_flag_id: str, *args: Any, **kwargs: Any) ->
...
+        ):
```
samsadsam, line 104: I guess the code block within this if statement should be:

```
result.user = targeting_context.user_id
if inspect.iscoroutinefunction(self._on_feature_evaluated):
    await self._on_feature_evaluated(result)
else:
    self._on_feature_evaluated(result)
```

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -53,50 +55,48 @@ def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
...
-                        and allocation.variant == evaluation_event.variant.name
-                        and allocation.percentile_to
-                    ):
-                        allocation_percentage += allocation.percentile_to - allocation.percentile_from
-                event["VariantAssignmentPercentage"] = str(allocation_percentage)
+    allocation_percentage = 0
+    if reason == VariantAssignmentReason.DEFAULT_WHEN_ENABLED and feature.allocation and feature.allocation.percentile:
+        for allocation in feature.allocation.percentile:
+            allocation_percentage += allocation.percentile_to - allocation.percentile_from
+
```
samsadsam, line 84: nit: remove this line

featuremanagement/_featuremanagerbase.py
```diff
@@ -225,31 +226,74 @@ def _build_targeting_context(self, args: Tuple[Any]) -> TargetingContext:
...
-            self._assign_variant(feature_flag, targeting_context, evaluation_event)
+        if not feature_flag.variants:
+            return
+
+        if not feature_flag.allocation:
+            evaluation_event.reason = (
+                VariantAssignmentReason.DEFAULT_WHEN_ENABLED
+                if evaluation_event.enabled
+                else VariantAssignmentReason.DEFAULT_WHEN_DISABLED
+            )
```
samsadsam, line 250: if allocation is `None`, I thought that the reason should be `None` too, unless I missed something

featuremanagement/_featuremanagerbase.py
```diff
@@ -149,56 +150,57 @@ def _is_targeted(context_id: str) -> float:
...
+                percentile_to: int = percentile_allocation.percentile_to
+                if (box == 100 and percentile_to == 100) or (
+                    percentile_allocation.percentile_from <= box < percentile_to
+                ):
                     evaluation_event.reason = VariantAssignmentReason.PERCENTILE
                     variant_name = percentile_allocation.variant
+                    break
+
         if not variant_name:
             FeatureManagerBase._check_default_enabled_variant(evaluation_event)
```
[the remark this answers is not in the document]
mrm9084: One issue with that is our style has a limit to the number of if/for statements we can have in a single method.

featuremanagement/_featuremanagerbase.py
```diff
@@ -149,56 +150,57 @@ def _is_targeted(context_id: str) -> float:
...
                     feature_flag, feature_flag.allocation.default_when_enabled
                 )
                 return
+
         evaluation_event.variant = self._variant_name_to_variant(feature_flag, variant_name)
-        if not feature_flag.variants:
-            return
-        FeatureManagerBase._check_variant_override(feature_flag.variants, variant_name, True, evaluation_event)
+        if feature_flag.variants:
+            FeatureManagerBase._check_variant_override(feature_flag.variants, variant_name, True, evaluation_event)
```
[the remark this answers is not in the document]
mrm9084: We have the name of the variant. We don't have the actual variant object we will return.

featuremanagement/_featuremanagerbase.py
mrm9084: I think this might be old code. Removed.

featuremanagement/_featuremanagerbase.py
mrm9084: It would be when `status_override` of the default disabled variant is `true`. `is_enabled` should never return true when a feature flag has `enabled=False` but we still need to assign variants, reasons, and telemetry, and such.

featuremanagement/_featuremanagerbase.py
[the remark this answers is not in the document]
samsadsam: adding the reason assignment directly and calling a function (`FeatureManagerBase._check_variant_override`) does not add a nesting level right?

featuremanagement/_featuremanagerbase.py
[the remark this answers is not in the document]
samsadsam: we have it in `evaluation_event.variant`

featuremanagement/_featuremanagerbase.py
mrm9084: @rossgrambo. I don't remember why we did this.

featuremanagement/_featuremanagerbase.py
samsadsam: Yes it's when `status_override=true`, I made a mistake. But seems here to me this line exists because there is a side effect to `self._assign_default_disabled_variant()` as it calls `self.__assign_variant_override()`.

If we did `evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED` at line 287, this line would not be necessary.

featuremanagement/_featuremanagerbase.py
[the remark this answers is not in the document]
mrm9084: Sorry confusing thing. One of these is a `VariantReference`, which is the object we use to figure out what to do. `Varaint` is the object we return to be used from `get_variant`.

featuremanagement/_featuremanagerbase.py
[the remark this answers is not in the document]
samsadsam: Sure, make sense. Resolving

featuremanagement/_featuremanagerbase.py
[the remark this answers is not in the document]
samsadsam: Why do we have both classes? They seem to only have one boolean (_status_override) as a difference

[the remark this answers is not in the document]
mrm9084: status override isn't returned to the customer. .Net has the same design.

featuremanagement/_featuremanagerbase.py
rossgrambo: Looks like it was changed in this PR- https://github.com/microsoft/FeatureManagement-Python/pull/24/files#diff-2a612ba07ea566bb99c5daad5ef16148ad7be8d42fdf61734f95ff96b43d73a5R318

I don't think I saw this change. Looking at it now- I don't think it makes sense. You handle this reason in the `_assign_default_disabled_variant` and `_assign_default_enabled_variant` methods.

featuremanagement/_featuremanagerbase.py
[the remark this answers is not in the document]
samsadsam: sure, thanks!

samsadsam (approved):

PR #51, opened 2024-12-11
Recurring time window filter

mrm9084:
# Description

Add the Recurring time window filter to python Feature Management.


featuremanagement/_time_window_filter/_recurrence_evaluator.py
```diff
@@ -0,0 +1,117 @@
...
+    first_day_of_first_week = start - timedelta(days=_get_passed_week_days(start.weekday(), pattern.first_day_of_week))
+
+    number_of_interval = (now - first_day_of_first_week).days // (interval * 7)
+    first_day_of_most_recent_occurring_week = first_day_of_first_week + timedelta(
+        days=number_of_interval * (interval * 7)
+    )
+    sorted_days_of_week = _sort_days_of_week(pattern.days_of_week, pattern.first_day_of_week)
+    max_day_offset = _get_passed_week_days(sorted_days_of_week[-1], pattern.first_day_of_week)
+    min_day_offset = _get_passed_week_days(sorted_days_of_week[0], pattern.first_day_of_week)
+    num_of_occurrences = number_of_interval * len(sorted_days_of_week) - sorted_days_of_week.index(start.weekday())
```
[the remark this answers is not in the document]
mrm9084: Pretty sure I have it correct here. https://github.com/microsoft/FeatureManagement-Dotnet/blob/325271bd622aa43d041b366c7ba2e9c30507e5a0/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs#L244, maybe it's the name causing an issue.

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,119 @@
...
+
+class RecurrencePattern:  # pylint: disable=too-few-public-methods
+    """
+    The recurrence pattern settings.
+    """
+
+    def __init__(self, pattern_data: Dict[str, Any]):
+        self.type = RecurrencePatternType.from_str(pattern_data.get("Type", "Daily"))
+        self.interval = pattern_data.get("Interval", 1)
+        self.days_of_week = pattern_data.get("DaysOfWeek", [])
```
zhiyuanliang-ms, line 74: How does it convert day of week string("Monday"）to index/number?

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class RecurrencePattern:  # pylint: disable=too-few-public-methods
+    """
+    The recurrence pattern settings.
+    """
+
+    def __init__(self, pattern_data: Dict[str, Any]):
+        self.type = RecurrencePatternType.from_str(pattern_data.get("Type", "Daily"))
+        self.interval = pattern_data.get("Interval", 1)
```
zhiyuanliang-ms, line 73: Interval can be non-positive or even not a number.

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    The recurrence range settings.
+    """
+
+    def __init__(self, range_data: Dict[str, Any]):
+        self.type = RecurrenceRangeType.from_str(range_data.get("Type", "NoEnd"))
+        if range_data.get("EndDate") and isinstance(range_data.get("EndDate"), str):
+            end_date_str = range_data.get("EndDate", "")
+            self.end_date = parsedate_to_datetime(end_date_str) if end_date_str else None
+        self.num_of_occurrences = range_data.get("NumberOfOccurrences")
```
zhiyuanliang-ms, line 88: Missed the default value. BTW, it is possible to get non number value or negative value.

zhiyuanliang-ms:
@mrm9084 Will you add testcases in this PR?

featuremanagement/_time_window_filter/_models.py
mrm9084: Isn't it always just a number? Otherwise, how do we deal with multiple languages? De we just assume English?

featuremanagement/_time_window_filter/_models.py
mrm9084: Is the default value 0?

zhiyuanliang-ms: https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs#L26

Default value is int.max

featuremanagement/_time_window_filter/_models.py
zhiyuanliang-ms: The schema is here: https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.TimeWindow.v2.0.0.schema.json

We assume English. We followed outlook schema: https://learn.microsoft.com/en-us/graph/api/resources/recurrencepattern?view=graph-rest-1.0

> how do we deal with multiple languages

All of our API and schema are English.

featuremanagement/_time_window_filter/_models.py
mrm9084: Python doesn't have a "Max Int" value. We could do something like `math.pow(2, 63) - 1`.

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,128 @@
...
+    days: list = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"]
+
+    def __init__(self, pattern_data: Dict[str, Any]):
+        self.type = RecurrencePatternType.from_str(pattern_data.get("Type", "Daily"))
+        self.interval = pattern_data.get("Interval", 1)
+        if self.interval <= 0:
+            raise ValueError("The interval must be greater than 0.")
+        days_of_week = pattern_data.get("DaysOfWeek", [])
+        for day in days_of_week:
+            self.days_of_week.append(self.days.index(day))
```
zhiyuanliang-ms, line 80: You didn't initialize self.days_of_week as `[]`

featuremanagement/_defaultfilters.py
```diff
@@ -31,6 +41,8 @@
 FEATURE_FILTER_NAME_KEY = "Name"
 IGNORE_CASE_KEY = "ignore_case"
 
+logger = logging.getLogger(__name__)
+
```
zhiyuanliang-ms, line 45: nit: extra empty line

featuremanagement/_defaultfilters.py
```diff
@@ -31,6 +41,8 @@
...
 
 class TargetingException(Exception):
```
zhiyuanliang-ms, line 47: Why is this class called "Targeting" exception? Shouldn't it be TimeWindowException?

featuremanagement/_defaultfilters.py
```diff
@@ -52,17 +64,45 @@ def evaluate(self, context: Mapping[Any, Any], **kwargs: Any) -> bool:
...
 
-        start_time = parsedate_to_datetime(start) if start else None
-        end_time = parsedate_to_datetime(end) if end else None
+        start_time: Optional[datetime] = parsedate_to_datetime(start) if start else None
+        end_time: Optional[datetime] = parsedate_to_datetime(end) if end else None
+
+        if recurrence:
+            if start_time and end_time:
+                settings = TimeWindowFilterSettings(start_time, end_time, recurrence)
+                return is_match(settings, current_time)
```
zhiyuanliang-ms, line 93: A better timewindow evaluation is 

``` python
if not start and not end:
    return False

if (start_time is None or start_time <= current_time) and (end_time is None or current_time < end_time):
    return True

if recurrence_data:
    recurrence = Recurrence(recurrence_data)
    settings = TimeWindowFilterSettings(start_time, end_time, recurrence)
    return is_match(settings, current_time)

return False
```

In `is_match`, you will validate settings (which may take some time), if the current time hits the first time window (specified by `start` and `end`), we can save the recurrence evaluation.

featuremanagement/_time_window_filter/_recurrence_evaluator.py
```diff
@@ -0,0 +1,121 @@
...
+    Check if the current time is within the time window filter settings.
+
+    :param TimeWindowFilterSettings settings: The settings for the time window filter.
+    :param datetime now: The current time.
+    :return: True if the current time is within the time window filter settings, otherwise False.
+    :rtype: bool
+    """
+    validate_settings(settings)
+
+    previous_occurrence = _get_previous_occurrence(settings, now)
```
zhiyuanliang-ms, line 23: How about returning False directly here `if now < start here`. In this way, we can save a function call (very slightly improve the performance).

featuremanagement/_time_window_filter/_recurrence_evaluator.py
```diff
@@ -0,0 +1,121 @@
...
+
+def _get_previous_occurrence(settings: TimeWindowFilterSettings, now: datetime) -> Optional[datetime]:
+    start = settings.start
+    if now < start:
+        return None
+
+    pattern_type = settings.recurrence.pattern.type
+    if pattern_type == RecurrencePatternType.DAILY:
+        occurrence_info = _get_daily_previous_occurrence(settings, now)
+    else:
```
zhiyuanliang-ms, line 39: You didn't validate pattern type. Here you assume the pattern type must be either DAILY or WEEKLY. But this is not guaranteed.

How about
``` python
if pattern_type == RecurrencePatternType.DAILY:
     occurrence_info = _get_daily_previous_occurrence(settings, now)
elif pattern_type == RecurrencePatternType.WEEKLY:
     occurrence_info = _get_weekly_previous_occurrence(settings, now)
else:
     throw
```

featuremanagement/_time_window_filter/_recurrence_evaluator.py
```diff
@@ -0,0 +1,121 @@
...
+        occurrence_info = _get_weekly_previous_occurrence(settings, now)
+
+    recurrence_range = settings.recurrence.range
+    range_type = recurrence_range.type
+    previous_occurrence = occurrence_info.previous_occurrence
+    end_date = recurrence_range.end_date
```
zhiyuanliang-ms, line 45: Is it a string or datetime type?

featuremanagement/_time_window_filter/_recurrence_evaluator.py
```diff
@@ -0,0 +1,121 @@
...
+    return OccurrenceInfo(most_recent_occurrence, num_of_occurrences)
+
+
+def _get_passed_week_days(current_day: int, first_day_of_week: int) -> int:
+    return (current_day - first_day_of_week + 7) % 7
+
+
+def _sort_days_of_week(days_of_week: List[int], first_day_of_week: int) -> List[int]:
+    sorted_days = sorted(days_of_week)
+    return sorted_days[sorted_days.index(first_day_of_week) :] + sorted_days[: sorted_days.index(first_day_of_week)]
```
zhiyuanliang-ms, line 121: What will happen if `first_day_of_week` is not in the list?

featuremanagement/_time_window_filter/_recurrence_validator.py
```diff
@@ -0,0 +1,162 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# -------------------------------------------------------------------------
+from datetime import datetime, timedelta
+from typing import List
+from ._models import RecurrencePatternType, RecurrenceRangeType, TimeWindowFilterSettings
+
+
+DAYS_PER_WEEK = 7
```
zhiyuanliang-ms, line 11: In other files, you used number 7 directly. So maybe we don't need this variable here.

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,128 @@
...
+        first_day_of_week = pattern_data.get("FirstDayOfWeek", "Sunday")
+        self.first_day_of_week = self.days.index(first_day_of_week) if first_day_of_week in self.days else 0
```
zhiyuanliang-ms, line 82:
``` python
from email.utils import parsedate_to_datetime

s = "Sun, 05 Jan 2025 20:00:00 GMT"
t = parsedate_to_datetime(s)
print(t.weekday())
```

This produced 6 on my laptop, I am using python 3.11.9.

first day of week should be Sunday by default. Sunday index is 6 and Monday is 0.

featuremanagement/_defaultfilters.py
mrm9084: Having two empty lines before the start of a class is part of the `black` formatting that we use.

featuremanagement/_time_window_filter/_recurrence_evaluator.py
mrm9084: This sort of sounds good, but I disagree. I would hate to have an invalid filter but only learn about it after it starts.

featuremanagement/_time_window_filter/_recurrence_evaluator.py
mrm9084: `datetime`, I convert it from `string` to `datetime` in the `RecurrenceRange` init.

featuremanagement/_time_window_filter/_recurrence_evaluator.py
mrm9084: It is guaranteed to be. In the `RecurrencePattern` init we validate that it is, and if it isn't we default to the index value of 0, which is Sunday.

featuremanagement/_time_window_filter/_recurrence_validator.py
mrm9084: I prefer the constant, I updated recurrence evaluator to use a constant too.

featuremanagement/_time_window_filter/_models.py
mrm9084: I've thought about this and I don't want to mess with the days of the week in python. So, what I think should be fine is that I update. `RecurrencePattern` to start on `Monday` which is correct in python, but still have the default first day of the week to be `"Sunday"` and as the input and default are both strings and we get the index from the list we should be fine. Thoughts?

featuremanagement/_time_window_filter/_recurrence_evaluator.py
zhiyuanliang-ms: In python Sunday index is 6, isn't it?
image
The `first_day_of_week` is Sunday (6) by default.
Let's say `days_of_week` is [0, 1] (Mon and Tue)
image
Won't `_sort_days_of_week` break as Sunday (6) is not in the list?

featuremanagement/_time_window_filter/_recurrence_validator.py
```diff
@@ -0,0 +1,164 @@
...
+        raise ValueError("Recurrence is required")
+
+    _validate_start_end_parameter(start, end)
+    _validate_recurrence_pattern(recurrence.pattern, start, end)
+    _validate_recurrence_range(recurrence.range, start)
+
+
+def _validate_start_end_parameter(start: datetime, end: datetime) -> None:
+    param_name = "end"
+    if end > start + timedelta(days=TEN_YEARS):
```
rossgrambo, line 40: @zhiyuanliang-ms why the 10 year limit?

tests/time_window_filter/recurrence_util.py
```diff
@@ -0,0 +1,48 @@
+from datetime import datetime, timedelta
```
rossgrambo, line 1: `featuremanagement/_time_window_filter/conversiontest.py` empty file was added

(github won't let me comment on that file directly)

rossgrambo (approved):

featuremanagement/_time_window_filter/_recurrence_validator.py
zhiyuanliang-ms: When I implemented recurring time window for .net, Jimmy said we should have a boundary. So I picked 10 years. I think it is not a big deal if we want to have a shorter one.

tests/time_window_filter/recurrence_util.py
mrm9084: Not sure how this got added. Removed.

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,139 @@
...
+            return RecurrenceRangeType.NUMBERED
+        raise ValueError(f"Invalid value: {value}")
+
+
+class RecurrencePattern:  # pylint: disable=too-few-public-methods
+    """
+    The recurrence pattern settings.
+    """
+
+    days: List[str] = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"]
```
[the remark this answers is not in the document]
mrm9084: I think this has changed a few times. The Python default is Monday = 0, but I think it is easier now to follow the app config default of Sunday = 0. Updating the comment and one other location.

zhiyuanliang-ms:
Why does it PR fail to build for python 3.8?

tests/time_window_filter/test_recurrence_evaluator.py
```diff
@@ -0,0 +1,279 @@
...
+from datetime import datetime
+import pytest
+from featuremanagement._time_window_filter._recurrence_evaluator import is_match
+from featuremanagement._time_window_filter._models import TimeWindowFilterSettings, Recurrence
+
+
+def test_is_match_within_time_window():
+    start = datetime(2025, 4, 7, 9, 0, 0)
+    end = datetime(2025, 4, 7, 17, 0, 0)
+    now = datetime(2025, 4, 7, 10, 0, 0)
```
zhiyuanliang-ms, line 15: Would you mind changing the `now` to 2025-4-8

zhiyuanliang-ms:
Can we add more testcases like this [PR](https://github.com/microsoft/FeatureManagement-JavaScript/pull/73/files#diff-4ca63141d1d877a7078cb592ecce0fd251b0940bfea8aa897d17cb5244d25431) did?

featuremanagement/_time_window_filter/_models.py
```diff
@@ -0,0 +1,138 @@
...
+    type: RecurrenceRangeType
+    end_date: Optional[datetime] = None
+
+    def __init__(self, range_data: Dict[str, Any]):
+        self.type = RecurrenceRangeType.from_str(range_data.get("Type", "NoEnd"))
+        if range_data.get("EndDate") and isinstance(range_data.get("EndDate"), str):
+            end_date_str = range_data.get("EndDate", "")
+            self.end_date = parsedate_to_datetime(end_date_str) if end_date_str else None
+        self.num_of_occurrences = range_data.get("NumberOfOccurrences", math.pow(2, 63) - 1)
+        if self.num_of_occurrences < 0:
```
zhiyuanliang-ms, line 103: num_of_occurrences should > 0

featuremanagement/_time_window_filter/_recurrence_validator.py
```diff
@@ -0,0 +1,164 @@
...
+def _sort_days_of_week(days_of_week: List[int], first_day_of_week: int) -> List[int]:
+    sorted_days = sorted(days_of_week)
+    if first_day_of_week in sorted_days:
+        return sorted_days[sorted_days.index(first_day_of_week) :] + sorted_days[: sorted_days.index(first_day_of_week)]
+    next_closet_day = first_day_of_week
+    for day in sorted_days:
+        if day > first_day_of_week:
+            next_closet_day = day
+            break
+    return sorted_days[sorted_days.index(next_closet_day) :] + sorted_days[: sorted_days.index(next_closet_day)]
```
zhiyuanliang-ms, line 164: image

The `_sort_days_of_week` method will break if the input is days_of_week: [0,1,2] and first_day_of_week: 6

But the expected result should be [0,1,2]

featuremanagement/_time_window_filter/_recurrence_evaluator.py
mrm9084: For my original reply, when I say 0 is the default, i.e. Sunday, That was prior to a commit, which I updated the list to start with Monday, which it does in .Net.

mrm9084: Also, if you look at the new sorted function, it fixes this by finding the closest day of the week to the first day of the week if the first day of the week isn't in the list.

featuremanagement/_time_window_filter/_models.py
mrm9084: Do we not allow for `num_of_occurrences` to be 0?

featuremanagement/_time_window_filter/_recurrence_validator.py
mrm9084: Fixed, and added this as a test.

mrm9084:
> Why does it PR fail to build for python 3.8?

I looked into it a bit, seems like a pylon bug in 3.8. I've added an empty init to the test folder which seems to have fixed it. Not sure why it wasn't an issue till now.

featuremanagement/_time_window_filter/_models.py
mrm9084: Updated.

tests/time_window_filter/test_recurrence_evaluator.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # Verify that the main method is_match correctly handles the scenario
+    assert is_match(settings, now) is False
+    assert is_match(settings, start) is True
+
+
+def test_weekly_recurrence_now_before_min_offset():
+    start = datetime(2025, 4, 9, 9, 0, 0)  # Monday
+    end = datetime(2025, 4, 9, 17, 0, 0)  # Monday
+    now = datetime(2025, 4, 16, 8, 0, 0)
```
zhiyuanliang-ms, line 262:
```suggestion
    now = datetime(2025, 4, 16, 8, 0, 0) # Saturday
```

tests/time_window_filter/test_recurrence_evaluator.py
```diff
@@ -0,0 +1,279 @@
...
+                "FirstDayOfWeek": "Monday",
+            },
+            "Range": {"Type": "Numbered", "NumberOfOccurrences": 4},
+        }
+    )
+
+    settings = TimeWindowFilterSettings(start=start, end=end, recurrence=recurrence)
+
+    # First occurrence should match
+    assert is_match(settings, datetime(2025, 4, 7, 10, 0, 0)) is True
```
zhiyuanliang-ms, line 214: Could add comment to indicate the day of week of these date?

tests/time_window_filter/test_recurrence_evaluator.py
```diff
@@ -0,0 +1,279 @@
...
+    assert is_match(settings, datetime(2025, 4, 8, 10, 0, 0)) is True
+
+    # Second week occurrence shouldn't match
+    assert is_match(settings, datetime(2025, 4, 14, 10, 0, 0)) is False
+    assert is_match(settings, datetime(2025, 4, 15, 10, 0, 0)) is False
+
+    assert is_match(settings, datetime(2025, 4, 7, 8, 0, 0)) is False
```
zhiyuanliang-ms, line 221: This line should be moved to the front

featuremanagement/_time_window_filter/_recurrence_evaluator.py
```diff
@@ -0,0 +1,125 @@
...
+    if now > first_day_of_most_recent_occurring_week + timedelta(days=DAYS_PER_WEEK):
+        num_of_occurrences += len(sorted_days_of_week)
+        most_recent_occurrence = first_day_of_most_recent_occurring_week + timedelta(days=max_day_offset)
+        return OccurrenceInfo(most_recent_occurrence, num_of_occurrences)
+
+    day_with_min_offset = first_day_of_most_recent_occurring_week + timedelta(days=min_day_offset)
+    if start > day_with_min_offset:
+        num_of_occurrences = 0
+        day_with_min_offset = start
+    if now < day_with_min_offset:
```
zhiyuanliang-ms, line 106: We don't guard that `start <= now` in this function. But I think it is ok

zhiyuanliang-ms (approved):

rossgrambo (approved):

PR #69, opened 2026-06-29
Update telemetry event dependency to OpenTelemetry LoggingHandler

yuanqu72:
**Overview**
This pull request modernizes the Azure Monitor telemetry integration by switching from the deprecated `azure-monitor-events-extension` to the OpenTelemetry logging handler. It updates the telemetry event sending logic, improves the handling of reserved log attributes, and refactors tests to match the new implementation. Dependency management and sample requirements are also updated for consistency.

**Telemetry integration updates:**

* Replaces usage of `azure-monitor-events-extension` with OpenTelemetry's `LoggingHandler` for sending telemetry events, introducing a dedicated logger (`_event_logger`) and initialization logic in `_send_telemetry.py`. This ensures telemetry events are sent in a more standard and extensible way.
* Updates the `track_event` and `publish_telemetry` functions to use the new logging-based approach, including logic to avoid conflicts with reserved log record attributes. [[1]](diffhunk://#diff-e599fecd01f487185b6b19104e39d7a20d310657896ddeee742af944b129945aL54-R93) [[2]](diffhunk://#diff-e599fecd01f487185b6b19104e39d7a20d310657896ddeee742af944b129945aL71-R102)
* Updates the `TargetingSpanProcessor` to check for OpenTelemetry logging support rather than the old extension.

**Dependency and configuration changes:**

* Updates the optional Azure Monitor dependency in `pyproject.toml` to require `opentelemetry-sdk` instead of `azure-monitor-events-extension`.
* Removes `azure-monitor-events-extension` from `samples/requirements.txt`.

**Testing improvements:**

* Refactors tests in `test_send_telemetry_appinsights.py` to patch the new `track_event` function instead of the old event extension, and adds a helper to extract event properties from mock calls. [[1]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401R16-R19) [[2]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401L43-R84) [[3]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401L70-R108) [[4]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401L90-R136) [[5]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401L133-R173) [[6]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401L163-R204) [[7]](diffhunk://#diff-47ef66c67d7a4d819410031b07b4557addbf59c7be9b8384fe18ce94b6171401L193-R237)
* Adds a new test to verify that reserved log record attributes (like `microsoft.custom_event.name`) are handled safely and not overwritten.

**Validation**
* Ran [quote-of-the-day-python sample](https://github.com/Azure-Samples/quote-of-the-day-python) locally. Changed the FeatureManagement-Python dependency to the local version using command:
```
pip install -e C:\Users\yuanqu\workplace\FeatureManagement-Python
```
Verified that *quota-of-the-day app* is still able to send telemetry to my testing app insights in the correct location.
image

### DIFF
--- a/featuremanagement/azuremonitor/_send_telemetry.py
+++ b/featuremanagement/azuremonitor/_send_telemetry.py
@@ -80,11 +80,16 @@ def track_event(event_name: str, user: str, event_properties: Optional[Dict[str,
         event_properties[TARGETING_ID] = user
 
     # Azure Monitor exporter maps this attribute to customEvent telemetry name.
-    custom_event_attributes = {
-        **event_properties,
-        "microsoft.custom_event.name": event_name,
-    }
-    _event_logger.info(event_name, extra=custom_event_attributes)
+    custom_event_attributes = {**event_properties, "microsoft.custom_event.name": event_name}
+
+    # logging raises KeyError if an `extra` key overwrites a built-in LogRecord attribute (e.g. "name", "message").
+    reserved = logging.makeLogRecord({}).__dict__
+    safe_attributes: Dict[str, Optional[str]] = {}
+    for key, value in custom_event_attributes.items():
+        safe_key = key if key not in reserved else f"telemetry.{key}"
+        safe_attributes[safe_key] = value
+
+    _event_logger.info(event_name, extra=safe_attributes)
 
 
 def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
--- a/featuremanagement/azuremonitor/_send_telemetry.py
+++ b/featuremanagement/azuremonitor/_send_telemetry.py
@@ -7,6 +7,7 @@
 
 import logging
 import inspect
+from logging import INFO
 from typing import Any, Callable, Dict, Optional
 from .._models import VariantAssignmentReason, EvaluationEvent, TargetingContext
 
@@ -17,7 +18,6 @@ _event_logger = logging.getLogger(__name__ + ".events")
 _event_logger.propagate = False
 
 try:
-    from logging import INFO
     from opentelemetry.sdk._logs import LoggingHandler
     from opentelemetry.sdk.trace import SpanProcessor
 
@@ -28,20 +28,17 @@ except ImportError:
     SpanProcessor = object  # type: ignore
 
 
-_EVENTS_LOGGER_INITIALIZED: list[bool] = []
+_EVENTS_LOGGER_INITIALIZED: bool = False
 
 
 def _initialize_event_logger() -> None:
+    global _EVENTS_LOGGER_INITIALIZED  # pylint: disable=global-statement
     if _EVENTS_LOGGER_INITIALIZED:
-        return
-
-    if not HAS_OPENTELEMETRY_LOGGING:
-        logger.warning("OpenTelemetry logging handler is not installed. Telemetry will not be sent.")
         return
 
     _event_logger.addHandler(LoggingHandler())
     _event_logger.setLevel(INFO)
-    _EVENTS_LOGGER_INITIALIZED.append(True)
+    _EVENTS_LOGGER_INITIALIZED = True
 
 
 FEATURE_NAME = "FeatureName"
@@ -54,6 +51,7 @@ DEFAULT_WHEN_ENABLED = "DefaultWhenEnabled"
 VERSION = "Version"
 VARIANT_ASSIGNMENT_PERCENTAGE = "VariantAssignmentPercentage"
 MICROSOFT_TARGETING_ID = "Microsoft.TargetingId"
+CUSTOM_EVENT_NAME = "microsoft.custom_event.name"
 
 EVENT_NAME = "FeatureEvaluation"
 
@@ -79,16 +77,11 @@ def track_event(event_name: str, user: str, event_properties: Optional[Dict[str,
         event_properties[TARGETING_ID] = user
 
     # Azure Monitor exporter maps this attribute to customEvent telemetry name.
-    custom_event_attributes = {**event_properties, "microsoft.custom_event.name": event_name}
-
-    # logging raises KeyError if an `extra` key overwrites a built-in LogRecord attribute (e.g. "name", "message").
-    reserved = logging.makeLogRecord({}).__dict__
-    safe_attributes: Dict[str, Optional[str]] = {}
-    for key, value in custom_event_attributes.items():
-        safe_key = key if key not in reserved else f"telemetry.{key}"
-        safe_attributes[safe_key] = value
-
-    _event_logger.info(event_name, extra=safe_attributes)
+    custom_event_attributes = {
+        **event_properties,
+        CUSTOM_EVENT_NAME: event_name,
+    }
+    _event_logger.info(event_name, extra=custom_event_attributes)
 
 
 def publish_telemetry(evaluation_event: EvaluationEvent) -> None:
@@ -166,7 +159,7 @@ class TargetingSpanProcessor(SpanProcessor):
         :param parent_context: The parent context of the span.
         """
         if not HAS_OPENTELEMETRY_LOGGING:
-            logger.warning("OpenTelemetry logging handler is not installed.")
+            logger.info("OpenTelemetry logging handler is not installed.")
             return
         if self._targeting_context_accessor and callable(self._targeting_context_accessor):
             if inspect.iscoroutinefunction(self._targeting_context_accessor):
--- a/featuremanagement/azuremonitor/_send_telemetry.py
+++ b/featuremanagement/azuremonitor/_send_telemetry.py
@@ -30,6 +30,21 @@ except ImportError:
     Span = object  # type: ignore
     Context = object  # type: ignore
 
+FEATURE_NAME = "FeatureName"
+ENABLED = "Enabled"
+TARGETING_ID = "TargetingId"
+VARIANT = "Variant"
+REASON = "VariantAssignmentReason"
+
+DEFAULT_WHEN_ENABLED = "DefaultWhenEnabled"
+VERSION = "Version"
+VARIANT_ASSIGNMENT_PERCENTAGE = "VariantAssignmentPercentage"
+MICROSOFT_TARGETING_ID = "Microsoft.TargetingId"
+AZURE_MONITOR_EVENT_NAME = "microsoft.custom_event.name"
+
+EVENT_NAME = "FeatureEvaluation"
+
+EVALUATION_EVENT_VERSION = "1.0.0"
 
 _EVENTS_LOGGER_INITIALIZED: bool = False
 
@@ -42,23 +57,6 @@ def _initialize_event_logger() -> None:
     _event_logger.addHandler(LoggingHandler())
     _event_logger.setLevel(INFO)
     _EVENTS_LOGGER_INITIALIZED = True
-
-
-FEATURE_NAME = "FeatureName"
-ENABLED = "Enabled"
-TARGETING_ID = "TargetingId"
-VARIANT = "Variant"
-REASON = "VariantAssignmentReason"
-
-DEFAULT_WHEN_ENABLED = "DefaultWhenEnabled"
-VERSION = "Version"
-VARIANT_ASSIGNMENT_PERCENTAGE = "VariantAssignmentPercentage"
-MICROSOFT_TARGETING_ID = "Microsoft.TargetingId"
-CUSTOM_EVENT_NAME = "microsoft.custom_event.name"
-
-EVENT_NAME = "FeatureEvaluation"
-
-EVALUATION_EVENT_VERSION = "1.0.0"
 
 
 def track_event(event_name: str, user: str, event_properties: Optional[Dict[str, Optional[str]]] = None) -> None:
@@ -82,7 +80,7 @@ def track_event(event_name: str, user: str, event_properties: Optional[Dict[str,
     # Azure Monitor exporter maps this attribute to customEvent telemetry name.
     custom_event_attributes = {
         **event_properties,
-        CUSTOM_EVENT_NAME: event_name,
+        AZURE_MONITOR_EVENT_NAME: event_name,
     }
     _event_logger.info(event_name, extra=custom_event_attributes)
 
--- a/featuremanagement/azuremonitor/_send_telemetry.py
+++ b/featuremanagement/azuremonitor/_send_telemetry.py
@@ -28,11 +28,10 @@ except ImportError:
     SpanProcessor = object  # type: ignore
 
 
-_events_logger_initialized = False
+_EVENTS_LOGGER_INITIALIZED: list[bool] = []
 
 def _initialize_event_logger() -> None:
-    global _events_logger_initialized
-    if _events_logger_initialized:
+    if _EVENTS_LOGGER_INITIALIZED:
         return
 
     if not HAS_OPENTELEMETRY_LOGGING:
@@ -43,7 +42,8 @@ def _initialize_event_logger() -> None:
 
     _event_logger.addHandler(LoggingHandler())
     _event_logger.setLevel(INFO)
-    _events_logger_initialized = True
+    _EVENTS_LOGGER_INITIALIZED.append(True)
+
 
 FEATURE_NAME = "FeatureName"
 ENABLED = "Enabled"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -58,10 +58,9 @@ Homepage = "https://github.com/microsoft/FeatureManagement-Python"
 Issues = "https://github.com/microsoft/FeatureManagement-Python/issues"
 
 [project.optional-dependencies]
-AzureMonitor = ["azure-monitor-events-extension<2.0.0"]
+AzureMonitor = ["azure-monitor-opentelemetry < 2.0.0"]
 test = [
     "azure-monitor-opentelemetry < 2.0.0",
-    "azure-monitor-events-extension < 2.0.0",
 ]
```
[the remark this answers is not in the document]
yuanqu72: Will modify as suggested.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -12,20 +12,39 @@
...
+_EVENTS_LOGGER_INITIALIZED: list[bool] = []
+
+
+def _initialize_event_logger() -> None:
+    if _EVENTS_LOGGER_INITIALIZED:
+        return
+
+    if not HAS_OPENTELEMETRY_LOGGING:
+        logger.warning(
+            "OpenTelemetry logging handler is not installed. Telemetry will not be sent to Application Insights."
```
mrm9084: We actually don't know if they are sending to application insights, we are just logging to open telemetry. They can send them anywhere.

pyproject.toml
```diff
@@ -58,10 +58,10 @@ Homepage = "https://github.com/microsoft/FeatureManagement-Python"
 Issues = "https://github.com/microsoft/FeatureManagement-Python/issues"
 
 [project.optional-dependencies]
-AzureMonitor = ["azure-monitor-events-extension<2.0.0"]
+AzureMonitor = ["azure-monitor-opentelemetry < 2.0.0", "opentelemetry-instrumentation-logging >= 0.40b0, < 1.0.0"]
```
mrm9084:
```suggestion
AzureMonitor = [
    "azure-monitor-opentelemetry < 2.0.0",
    "opentelemetry-instrumentation-logging >= 0.40b0, < 1.0.0"
]
```

pyproject.toml
```diff
@@ -58,10 +58,10 @@ Homepage = "https://github.com/microsoft/FeatureManagement-Python"
...
 test = [
     "azure-monitor-opentelemetry < 2.0.0",
-    "azure-monitor-events-extension < 2.0.0",
+    "opentelemetry-instrumentation-logging >= 0.40b0, < 1.0.0",
```
mrm9084: We don't support betas as a rule. Is this not GAed?

mrm9084 (changes requested):

pyproject.toml
mrm9084: Also, is this a new dependency? It would break all existing customers.

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Good point. Will update the warning message.

pyproject.toml
yuanqu72: Updated code to depend on `telemetry_sdk`.

pyproject.toml
yuanqu72: Updated code to depend on telemetry_sdk.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,15 +68,27 @@ def track_event(event_name: str, user: str, event_properties: Optional[Dict[str,
...
 
-    azure_monitor_track_event(event_name, event_properties)
+    # Azure Monitor exporter maps this attribute to customEvent telemetry name.
+    custom_event_attributes = {**event_properties, "microsoft.custom_event.name": event_name}
+
+    # logging raises KeyError if an `extra` key overwrites a built-in LogRecord attribute (e.g. "name", "message").
+    reserved = logging.makeLogRecord({}).__dict__
+    safe_attributes: Dict[str, Optional[str]] = {}
+    for key, value in custom_event_attributes.items():
+        safe_key = key if key not in reserved else f"telemetry.{key}"
```
mrm9084: We should probably not just ignore reserved words, the customer is expecting us to log telemetry, we can't just ignore it.

tests/test_send_telemetry_appinsights.py
```diff
@@ -40,23 +44,44 @@ def test_send_telemetry_appinsights(self):
...
+            assert event_properties["Variant"] == "TestVariant"
+            assert event_properties["ETag"] == "cmwBRcIAq1jUyKL3Kj8bvf9jtxBrFg-R-ayExStMC90"
+            assert event_properties["FeatureFlagReference"] == "fake-store-uri/kv/.appconfig.featureflag/TestFeature"
+            assert event_properties["FeatureFlagId"] == "fake-feature-flag-id"
+
+    def test_track_event_preserves_reserved_custom_event_name(self):
+        with (
+            patch("featuremanagement.azuremonitor._send_telemetry._initialize_event_logger"),
+            patch("featuremanagement.azuremonitor._send_telemetry._event_logger.info") as mock_event_logger_info,
+            patch("featuremanagement.azuremonitor._send_telemetry.HAS_OPENTELEMETRY_LOGGING", True),
```
mrm9084: Do we need to patch this? This should be a test dependency. If anything, we'd patch it to false, so we can check our negative code path.

pyproject.toml
```diff
@@ -58,10 +58,9 @@ Homepage = "https://github.com/microsoft/FeatureManagement-Python"
 Issues = "https://github.com/microsoft/FeatureManagement-Python/issues"
 
 [project.optional-dependencies]
-AzureMonitor = ["azure-monitor-events-extension<2.0.0"]
+AzureMonitor = ["opentelemetry-sdk >= 1.20.0, < 2.0.0"]
```
mrm9084: I think this should also be a test dependency.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,15 +68,27 @@ def track_event(event_name: str, user: str, event_properties: Optional[Dict[str,
...
+    _initialize_event_logger()
+
     event_properties = event_properties or {}
 
     if user:
         event_properties[TARGETING_ID] = user
 
-    azure_monitor_track_event(event_name, event_properties)
+    # Azure Monitor exporter maps this attribute to customEvent telemetry name.
+    custom_event_attributes = {**event_properties, "microsoft.custom_event.name": event_name}
```
mrm9084: We might want to chat about if this should be changed so that everything else is in its own file and just the Azure Monitor stuff is done here.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -129,15 +158,15 @@ def __init__(self, **kwargs: Any) -> None:
...
         """
         Attaches the targeting ID to the span and baggage when a new span is started.
 
         :param Span span: The span that was started.
         :param parent_context: The parent context of the span.
         """
-        if not HAS_AZURE_MONITOR_EVENTS_EXTENSION:
-            logger.warning("Azure Monitor Events Extension is not installed.")
+        if not HAS_OPENTELEMETRY_LOGGING:
+            logger.warning("OpenTelemetry logging handler is not installed.")
```
mrm9084: I know I wrote it, but I agree with copilot this should be info not warning.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -12,20 +12,37 @@
...
 
+# Set up a separate logger for feature-evaluation telemetry events to avoid propagating to the root logger
+_event_logger = logging.getLogger(__name__ + ".events")
+_event_logger.propagate = False
+
 try:
-    from azure.monitor.events.extension import track_event as azure_monitor_track_event  # type: ignore
-    from opentelemetry.context.context import Context
-    from opentelemetry.sdk.trace import Span, SpanProcessor
+    from logging import INFO
```
mrm9084: This shouldn't be in the try block.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -12,20 +12,37 @@
...
+    from opentelemetry.sdk._logs import LoggingHandler
```
mrm9084, line 21: We can't import from private methods, I'm not should why this isn't being flaged.

mrm9084 (changes requested):

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -51,15 +68,27 @@ def track_event(event_name: str, user: str, event_properties: Optional[Dict[str,
...
 
     if user:
         event_properties[TARGETING_ID] = user
 
-    azure_monitor_track_event(event_name, event_properties)
+    # Azure Monitor exporter maps this attribute to customEvent telemetry name.
+    custom_event_attributes = {**event_properties, "microsoft.custom_event.name": event_name}
+
+    # logging raises KeyError if an `extra` key overwrites a built-in LogRecord attribute (e.g. "name", "message").
+    reserved = logging.makeLogRecord({}).__dict__
```
mrm9084: We should do this during the initialize there is no reason to do it every time.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -12,20 +12,37 @@
...
-        "azure-monitor-events-extension is not installed. Telemetry will not be sent to Application Insights."
-    )
+    HAS_OPENTELEMETRY_LOGGING = False
+    LoggingHandler = object  # type: ignore
     SpanProcessor = object  # type: ignore
-    Span = object  # type: ignore
-    Context = object  # type: ignore
+
+
+_EVENTS_LOGGER_INITIALIZED: list[bool] = []
```
mrm9084: Why is this a list? It only ever is one item.

featuremanagement/azuremonitor/_send_telemetry.py
mrm9084: `"microsoft.custom_event.name"` should be a constant.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -12,20 +12,37 @@
...
+
+
+def _initialize_event_logger() -> None:
+    if _EVENTS_LOGGER_INITIALIZED:
+        return
+
+    if not HAS_OPENTELEMETRY_LOGGING:
+        logger.warning("OpenTelemetry logging handler is not installed. Telemetry will not be sent.")
+        return
```
mrm9084: This is only ever called via `track_event` which already makes this check.

mrm9084 (changes requested):

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: This was carried over from existing implementation: https://github.com/microsoft/FeatureManagement-Python/blob/cc2c47fb23e5c0cda46298449398ddca64e5ca19/featuremanagement/azuremonitor/_send_telemetry.py#L20 I will move it out from the try-except block.

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: This is a pattern used across Azure Monitor and Opentelemetry packages. Even the previous dependency of `azure_monitor_track_event` uses the `sdk_logs` method: https://github.com/microsoft/ApplicationInsights-Python/blob/7e1ec03837b7a3b7e7ccaa395ea231e44e4f846e/azure-monitor-events-extension/azure/monitor/events/extension/_events.py#L6 I think we can use a similar pattern here?

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: This was modified based on one of copilot comments earlier. But I do agree that this should be one boolean value instead of a list. I will update it.

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: This is indeed a redundancy check which the code will never reach here. Will Remove it.

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Make sense. Will update the code.

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Will remove this line. Preservingthe previous usage makes more sense: https://github.com/microsoft/FeatureManagement-Python/blob/main/featuremanagement/azuremonitor/_send_telemetry.py#L62

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Agree. I will preserve the previous usage: https://github.com/microsoft/FeatureManagement-Python/blob/main/featuremanagement/azuremonitor/_send_telemetry.py#L62

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Will update to "info"

tests/test_send_telemetry_appinsights.py
yuanqu72: Validated that current test dependency  `azure-monitor-opentelemetry` will install `opentelemetry_sdk`: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-opentelemetry/setup.py#L87. We can remove this patch.

pyproject.toml
yuanqu72: Validated that current test dependency  `azure-monitor-opentelemetry` will install `opentelemetry_sdk`: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-opentelemetry/setup.py#L87. Therefore we do not need this as test dependency.

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -7,26 +7,43 @@
...
+
+
+def _initialize_event_logger() -> None:
+    global _EVENTS_LOGGER_INITIALIZED  # pylint: disable=global-statement
+    if _EVENTS_LOGGER_INITIALIZED:
+        return
+
+    _event_logger.addHandler(LoggingHandler())
+    _event_logger.setLevel(INFO)
+    _EVENTS_LOGGER_INITIALIZED = True
```
mrm9084: It feels weird to have this before all our constants

featuremanagement/azuremonitor/_send_telemetry.py
```diff
@@ -37,6 +54,7 @@
 VERSION = "Version"
 VARIANT_ASSIGNMENT_PERCENTAGE = "VariantAssignmentPercentage"
 MICROSOFT_TARGETING_ID = "Microsoft.TargetingId"
+CUSTOM_EVENT_NAME = "microsoft.custom_event.name"
```
mrm9084:
```suggestion
AZURE_MONITOR_EVENT_NAME = "microsoft.custom_event.name"
```

Thoughts?

tests/test_send_telemetry_appinsights.py
```diff
@@ -40,23 +44,43 @@ def test_send_telemetry_appinsights(self):
         evaluation_event.variant = variant
         evaluation_event.reason = VariantAssignmentReason.DEFAULT_WHEN_DISABLED
 
-        with patch("featuremanagement.azuremonitor._send_telemetry.azure_monitor_track_event") as mock_track_event:
+        with patch("featuremanagement.azuremonitor._send_telemetry.track_event") as mock_track_event:
```
mrm9084: This actually changes the test quite a bit it changes it from us testing what we are publishing to patching our own methods. Shouldn't we be patching this:

`_event_logger.info(event_name, extra=custom_event_attributes)`

mrm9084 (changes requested):

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Will move it after the constants definition.

featuremanagement/azuremonitor/_send_telemetry.py
yuanqu72: Will rename it locally and send another commit.

tests/test_send_telemetry_appinsights.py
yuanqu72: Agree, this is not what we intend to test. We do want to call `track_event` and only patch `_event_logger.info(event_name, extra=custom_event_attributes)`. Will update the unit tests.

mrm9084 (approved):

